PR #19, opened 2023-07-01
fix licence to pass opam lint

zapashcanon:


filipeom (approved):

filipeom:
Thank you so much! I was trying to remove that warning for a while

PR #20, opened 2023-07-11
functorize mappings

zapashcanon:
Hi,

I functorized the Z3 mappings. This is to prepare the upcoming PR adding a colibri2 backend. :)


filipeom (approved):
It looks great! Thanks!!

PR #21, opened 2023-07-11
More flexible dependency constraints

chambart:
Those versions of core and Z3 also works


filipeom (approved):

filipeom:
Thanks Pierre! :smiley:

PR #29, opened 2023-10-08
remove base/core

zapashcanon:
Hi,

As I told you a few weeks ago. :)

It went smoothly. One thing I'd like you to confirm me is that it's OK to use `Hashtbl.hash` as a hash function in `symbol.ml` ?

Also, I did not remove the dependency in `bin/`.

### DIFF
--- a/dune-project
+++ b/dune-project
@@ -22,8 +22,6 @@
    dune
    (ocaml (>= "4.14.0"))
    (z3 (and (>= "4.11.2") (< "4.13")))
-   core
-   core_unix
    menhir
    ppx_inline_test
    ppx_expect))
--- a/encoding.opam
+++ b/encoding.opam
@@ -11,8 +11,6 @@ depends: [
   "dune" {>= "3.0"}
   "ocaml" {>= "4.14.0"}
   "z3" {>= "4.11.2" & < "4.13"}
-  "core"
-  "core_unix"
   "menhir"
   "ppx_inline_test"
   "ppx_expect"
--- a/lib/ast.ml
+++ b/lib/ast.ml
@@ -1,5 +1,3 @@
-open Core
-
 type t =
   | Declare of Symbol.t
   | Assert of Expression.t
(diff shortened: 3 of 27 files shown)


filipeom:
Hi!

> It went smoothly. One thing I'd like you to confirm me is that it's OK to use Hashtbl.hash as a hash function in symbol.ml ?

Great! And yes, it is indeed Ok 

> Also, I did not remove the dependency in bin/

This, should be ok as well as the files in `bin/` will not be installed. If this changes in the future, I think we integrate Hichem's change to `Cmdliner` https://github.com/hra687261/encoding/commit/90fb8cb9f7926048513ca427d6b8ca410e986cf6

zapashcanon:
Oh but actually, the hash function is not needed anymore. It's because at some point I went with a `module Symboltbl = Hashtbl.Make (Symbol)` but then decided it was not worthwhile. I'll remove it.

filipeom:
> Oh but actually, the hash function is not needed anymore. It's because at some point I went with a `module Symboltbl = Hashtbl.Make (Symbol)` but then decided it was not worthwhile. I'll remove it.

That's correct. One thing we should also remove is the `[@@deriving compare, sexp, hash]` from the files:

- `types.ml`
- `bvOp.ml`
- `num.ml`
- `strOp.ml`
- `intOp.ml`
- `floatOp.ml`
- `boolOp.ml`

zapashcanon:
Indeed, good catch ! The compiler isn't telling anything about those of course... I removed them. :)

filipeom:
Thanks! Do you think I should remove `ppx_inline_test` and `ppx_expect` as depends? They depend on `base` and the unit tests aren't really testing anything

zapashcanon:
You could mark them to be test dependencies ? Something like this in `dune-project`: `(ppx_inline_test :with-test)`.

Also I'm not sure you can set `ppx_expect` as a test dependency only or if you're using it in the library directly and it may refuse to compile if not present.

Anyway, if you prefer to remove them, it's fine with me. :)

filipeom:
> You could mark them to be test dependencies ? Something like this in dune-project: (ppx_inline_test :with-test). 

Good idea! I'll to do this in the future and remove ppx_expect to reduce the time it takes to install dependencies

zapashcanon:
Fix #23.

PR #33, opened 2023-10-21
Adds Concat and Extract as operators for bit vector values

joaomhmpereira:


filipeom:
Looks good! Thanks Joao! Won't merge now as it will break owi

PR #36, opened 2023-11-18
Major expression and type rewrite

filipeom:
# First Encoding Version

The major rewrite of the expression and type modules marks the initial release of the encoding. Here, we aim to simplify the construction of arbitrary expressions without explicitly annotating each operator with its type. By constructing expressions in this manner, we can more easily apply generic simplification rules for the same operator across multiple SMT theories. Furthermore, I believe this approach could enhance caching and, perhaps, maintainability.

## New features

Furthermore, in this version we introduce:

- `I8`: Support for bitvectors with 8 bits


lib/expr.ml
```diff
@@ -0,0 +1,280 @@
...
+    | Relop (_, e1, e2) -> symbols e1 @ symbols e2
+    | Cvtop (_, e) -> symbols e
+    | Symbol s -> [ s ]
+    | Extract (e, _, _) -> symbols e
+    | Concat (e1, e2) -> symbols e1 @ symbols e2
+  in
+  List.fold_left
+    (fun accum x -> if List.mem x accum then accum else x :: accum)
+    []
+    (List.concat_map symbols e)
```
zapashcanon, line 61: The complexity of this function can be made much lower: in symbols, you could avoid concatenating lists which is expensive, in the `fold_left` call, `List.mem` on every element is making the whole thing quadratic. I would suggest something like:

```ocaml
let get_symbols e =
  let tbl = Hashtbl.create 64 in
  let rec symbols e = 
    match e.e with
    | Val _ -> ()
    | Ptr (_, offset) -> symbols offset
    | Unop (_, e1) -> symbols e1
    | Binop (_, e1, e2) -> symbols e1; symbols e2
    | Triop (_, e1, e2, e3) -> symbols e1; symbols e2; symbols e3
    | Relop (_, e1, e2) -> symbols e1; symbols e2
    | Cvtop (_, e) -> symbols e
    | Symbol s -> Hashtbl.replace tbl s ()
    | Extract (e, _, _) -> symbols e
    | Concat (e1, e2) -> symbols e1; symbols e2
  in
  List.iter symbols e;
  Hashtbl.fold (fun k () acc -> k::acc) tbl []
``` 
  (untested)

lib/expr.ml
```diff
@@ -0,0 +1,280 @@
...
+    | Concat (e1, e2) -> fprintf fmt "(++ %a %a)" pp e1 pp e2
+
+  let pp_list fmt (es : t list) =
+    Format.pp_print_list ~pp_sep:Format.pp_print_space pp fmt es
+
+  let pp_smt fmt (es : t list) : unit =
+    let pp_symbols fmt syms =
+      Format.pp_print_list ~pp_sep:Format.pp_print_newline
+        (fun fmt sym ->
+          let x = Symbol.to_string sym in
```
zapashcanon, line 105: We don't have a `Symbol.pp` yet ?

lib/expr.ml
```diff
@@ -0,0 +1,280 @@
...
+    | _ -> Relop (op, e1, e2) )
+  | Ptr (b1, os1), Ptr (b2, os2) -> (
+    let v i = { ty = Ty_bitv S32; e = Val (Num (I32 i)) } in
+    match op with
+    | Eq -> if b1 = b2 then Relop (Eq, os1, os2) else Val (Bool false)
+    | Ne -> if b1 = b2 then Relop (Ne, os1, os2) else Val (Bool false)
+    | LtU -> if b1 = b2 then Relop (LtU, os1, os2) else Relop (LtU, v b1, v b2)
+    | LeU -> if b1 = b2 then Relop (LeU, os1, os2) else Relop (LeU, v b1, v b2)
+    | GtU -> if b1 = b2 then Relop (GtU, os1, os2) else Relop (GtU, v b1, v b2)
+    | GeU -> if b1 = b2 then Relop (GeU, os1, os2) else Relop (GeU, v b1, v b2)
```
zapashcanon, line 190:
```ocaml
| LtU | LeU | GtU | GeU as op -> if b1 = b2 then Relop (op, os1, os2) else Relop (op, v b1, v b2)
```

?

lib/expr.ml
filipeom: Ah yes! Been meaning to optimize this function, this is perfect thanks!! :smiley:

lib/ty.ml
```diff
@@ -0,0 +1,188 @@
...
+  | TruncUF32
+  | TruncSF64
+  | TruncUF64
+  | WrapI64
+  | ExtendSI32
+  | ExtendUI32
+
+let pp_unop fmt (op : unop) =
+  match op with
+  | Neg -> Format.pp_print_string fmt "neg"
```
zapashcanon, line 96: Maybe putting something like:

```ocaml
open Format
let pp_string = pp_print_string
let pp = fprintf
```

at the top would wake everything that follows less dense

lib/expr.ml
filipeom: We do, I'll use it here instead :smiley:

lib/eval_numeric.ml
```diff
@@ -19,7 +19,9 @@ module I32Op = struct
   let to_value i : Num.t = I32 i
 
   let of_value n v : t =
-    of_arg (fun v -> match v with I32 i -> i | _ -> raise (Num `I32Type)) n v
+    of_arg
+      (fun v -> match v with I32 i -> i | _ -> raise (Num (Ty_bitv S32)))
```
zapashcanon, line 23:
```ocaml
of_arg (function I32 i -> i | _ -> raise ...)
```

:-)

lib/expr.ml
filipeom: Much better, thanks! :grin:

lib/eval_numeric.ml
```diff
@@ -96,7 +103,9 @@ module I64Op = struct
   let to_value i : Num.t = I64 i
 
   let of_value n v : int64 =
-    of_arg (fun v -> match v with I64 i -> i | _ -> raise (Num `I64Type)) n v
+    of_arg
+      (fun v -> match v with I64 i -> i | _ -> raise (Num (Ty_bitv S64)))
```
zapashcanon, line 107: same here

lib/eval_numeric.ml
```diff
@@ -173,12 +187,12 @@ module F32Op = struct
   let to_value f : Num.t = F32 f
 
   let of_value =
-    of_arg (fun v -> match v with F32 f -> f | _ -> raise (Num `F32Type))
+    of_arg (fun v -> match v with F32 f -> f | _ -> raise (Num (Ty_fp S32)))
```
zapashcanon, line 190: same here

lib/eval_numeric.ml
```diff
@@ -223,12 +239,12 @@ module F64Op = struct
   let to_value f : Num.t = F64 f
 
   let of_value =
-    of_arg (fun v -> match v with F64 f -> f | _ -> raise (Num `F64Type))
+    of_arg (fun v -> match v with F64 f -> f | _ -> raise (Num (Ty_fp S64)))
```
zapashcanon, line 242: same here

lib/parser.mly
```diff
@@ -43,14 +44,18 @@ let stmt :=
...
+  | c = spec_constant; { { ty = Value.type_of c; e = Val c } }
+  | LPAREN; (ty, op) = UNARY; e = s_expr; RPAREN;
+    { { ty; e = Unop (op, e) } }
+  | LPAREN; (ty, op) = BINARY; e1 = s_expr; e2 = s_expr; RPAREN;
+    { { ty; e = Binop (op, e1, e2) } }
+  | LPAREN; (ty, op) = TERNARY; e1 = s_expr; e2 = s_expr; e3 = s_expr; RPAREN;
+    { { ty; e = Triop (op, e1, e2, e3) } }
+  | LPAREN; (ty, op) = CVTOP; e = s_expr; RPAREN;
+    { { ty; e = Cvtop (op, e) } }
+  | LPAREN; (ty, op) = RELOP; e1 = s_expr; e2 = s_expr; RPAREN;
```
zapashcanon, line 57: It should be possible to move all the `LPAREN; ...; RPAREN;` to their own rule `paren_op` for instance and have a single `LPAREN; op = paren_op; RPAREN`.

lib/z3_mappings.ml
```diff
@@ -265,160 +279,146 @@ module Fresh = struct
...
-            fun e -> Boolean.mk_ite ctx e (encode_val 1L) (encode_val 0L)
-          | WrapI64 -> assert false
+          match sz with
+          | Ty.S32 -> (
+            match op with
+            | WrapI64 -> BitVector.mk_extract ctx 31 0
+            | TruncSF32 -> fun f -> FloatingPoint.mk_to_sbv ctx rtz f 32
+            | TruncUF32 -> fun f -> FloatingPoint.mk_to_ubv ctx rtz f 32
+            | TruncSF64 -> fun f -> FloatingPoint.mk_to_sbv ctx rtz f 32
+            | TruncUF64 -> fun f -> FloatingPoint.mk_to_ubv ctx rtz f 32
```
zapashcanon, line 373: These four cases could be merged in only two ?

lib/z3_mappings.ml
```diff
@@ -265,160 +279,146 @@ module Fresh = struct
...
+            | ExtendSI32 | ExtendUI32 | _ -> assert false )
+          | Ty.S64 -> (
+            match op with
+            | ExtendSI32 -> BitVector.mk_sign_ext ctx 32
+            | ExtendUI32 -> BitVector.mk_zero_ext ctx 32
+            (* rounding towards zero (aka truncation) *)
+            | TruncSF32 -> fun f -> FloatingPoint.mk_to_sbv ctx rtz f 64
+            | TruncUF32 -> fun f -> FloatingPoint.mk_to_ubv ctx rtz f 64
+            | TruncSF64 -> fun f -> FloatingPoint.mk_to_sbv ctx rtz f 64
+            | TruncUF64 -> fun f -> FloatingPoint.mk_to_ubv ctx rtz f 64
```
zapashcanon, line 388: These four cases could be merged in only two ?

lib/z3_mappings.ml
```diff
@@ -441,202 +444,169 @@ module Fresh = struct
...
+            | DemoteF64 ->
+              fun bv -> FloatingPoint.mk_to_fp_float ctx rne bv fp32_sort
+            | ConvertSI32 ->
+              fun bv -> FloatingPoint.mk_to_fp_signed ctx rne bv fp32_sort
+            | ConvertUI32 ->
+              fun bv -> FloatingPoint.mk_to_fp_unsigned ctx rne bv fp32_sort
+            | ConvertSI64 ->
+              fun bv -> FloatingPoint.mk_to_fp_signed ctx rne bv fp32_sort
+            | ConvertUI64 ->
+              fun bv -> FloatingPoint.mk_to_fp_unsigned ctx rne bv fp32_sort
```
zapashcanon, line 479: These four cases could be merged in only two ?

lib/z3_mappings.ml
```diff
@@ -441,202 +444,169 @@ module Fresh = struct
...
+            | PromoteF32 ->
+              fun bv -> FloatingPoint.mk_to_fp_float ctx rne bv fp64_sort
+            | ConvertSI32 ->
+              fun bv -> FloatingPoint.mk_to_fp_signed ctx rne bv fp64_sort
+            | ConvertUI32 ->
+              fun bv -> FloatingPoint.mk_to_fp_unsigned ctx rne bv fp64_sort
+            | ConvertSI64 ->
+              fun bv -> FloatingPoint.mk_to_fp_signed ctx rne bv fp64_sort
+            | ConvertUI64 ->
+              fun bv -> FloatingPoint.mk_to_fp_unsigned ctx rne bv fp64_sort
```
zapashcanon, line 496: These four cases could be merged in only two ?

lib/parser.mly
filipeom: Oh yes, nice thinking! It makes the parsing rules more concise :grin:

lib/z3_mappings.ml
filipeom: Thanks, yes we can merge all of these! :grinning:

lib/ty.ml
filipeom: Indeed, it looks much better with this

lib/value.ml
```diff
@@ -1,34 +1,38 @@
...
 let pp fmt (v : t) =
   match v with
-  | Int x -> Format.pp_print_string fmt @@ Int.to_string x
-  | Real x -> Format.pp_print_string fmt @@ Float.to_string x
-  | Bool x -> Format.pp_print_string fmt @@ Bool.to_string x
+  | True -> pp_string fmt "true"
+  | False -> pp_string fmt "false"
+  | Int x -> pp_string fmt @@ Int.to_string x
+  | Real x -> pp_string fmt @@ Float.to_string x
   | Num x -> Format.fprintf fmt "%a" Num.pp x
```
zapashcanon, line 36: This can be `Num.pp fmt x`

filipeom: Fixed here :smile:  https://github.com/formalsec/encoding/commit/2f6ae24337f37b59f0bdd347681b65cb8e218f0a

PR #37, opened 2023-11-25
Nary operators

joaomhmpereira:
Draft

### DIFF
--- a/lib/expr.ml
+++ b/lib/expr.ml
@@ -102,6 +102,22 @@ let rec ty (hte : t) : Ty.t =
     | Ty_bitv n1, Ty_bitv n2 -> Ty_bitv (n1 + n2)
     | t1, t2 -> Log.err "Invalid concat of (%a) with (%a)" Ty.pp t1 Ty.pp t2 )
 
+let rec is_symbolic (v : t) : bool =
+  match view v with
+  | Val _ -> false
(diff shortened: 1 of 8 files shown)


lib/expr.ml
```diff
@@ -52,6 +52,20 @@ let rec equal (e1 : t) (e2 : t) : bool =
     | Concat (e1, e3), Concat (e2, e4) -> equal e1 e2 && equal e3 e4
     | _ -> false
 
+let rec is_symbolic (v : t) : bool =
+    match v.e with
+    | Val _ -> false
+    | Symbol _ -> true
+    | Unop (_, v) -> is_symbolic v
+    | Binop (_, v1, v2) -> is_symbolic v1 || is_symbolic v2
+    | Triop (_, v1, v2, v3) -> List.exists is_symbolic [ v1; v2; v3 ]
```
zapashcanon, line 61: It doesn't matter but:

```ocaml
| Triop (_, v1, v2, v3) -> Array.exists is_symbolic [| v1; v2; v3 |]
```

? :D
```

filipeom: Yes, this should be better :smiley:

filipeom:
Merging for now so that we can progress on formalsec/ECMA-SL#1.

Thanks!

PR #39, opened 2023-11-27
Hash-consing

filipeom:
https://www.lri.fr/~filliatr/ftp/publis/hash-consing2.pdf

### DIFF
--- a/lib/expr.ml
+++ b/lib/expr.ml
@@ -283,14 +283,41 @@ let rec simplify ?(extract = true) ({ ty; e } as expr : t) : t =
   | _ -> expr
 
 module Bitv = struct
-  let ty_of_cast (type a) (c : a Ty.cast) : Ty.t =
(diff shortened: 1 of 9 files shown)


lib/expr.ml
```diff
@@ -1,46 +1,65 @@
...
+      | Extract (e1, h1, l1), Extract (e2, h2, l2) ->
+        e1 == e2 && h1 = h2 && l1 = l2
+      | Concat (e1, e3), Concat (e2, e4) -> e1 == e2 && e3 == e4
+      | _ -> false
+
+  let hash = Hashtbl.hash
+end
+
+include M
+module H = Hashtbl.Make (M)
```
zapashcanon, line 45: It's probably better to use `Ephemeron.K1.Make`. At the time of that paper, ephemerons were not part of the stdlib, so they had to redo it from scratch, but now it's easy. :)

filipeom: Will do! Thanks for the suggestion :smile:

PR #43, opened 2023-11-30
Smtlib lexer, parser, and ast

filipeom:
Naive implementation of smt-lib v2.6


filipeom:
Needs more work https://github.com/filipeom/smtml/tree/smtlib

PR #44, opened 2023-12-07
use menhir short syntax

zapashcanon:
Hi @filipeom,

I used some syntax magic provided by menhir. :)

Also, any reason why you decided not to re-use [dolmen](https://github.com/gbury/dolmen) ?

### DIFF
--- a/lib/smtlib_lexer.mll
+++ b/lib/smtlib_lexer.mll
@@ -5,7 +5,7 @@ open Smtlib_parser
 exception SyntaxError of string
 
 let keywords =
-  let tbl = Hashtbl.create 10 in
+  let tbl = Hashtbl.create 16 in
   Array.iter
     (fun (k, v) -> Hashtbl.add tbl k v)
     [| (":sorts", SORTS)
@@ -22,7 +22,7 @@ let keywords =
   tbl
 
 let reserved =
-  let tbl = Hashtbl.create 45 in
+  let tbl = Hashtbl.create 64 in
   Array.iter
     (fun (k, v) -> Hashtbl.add tbl k v)
     [|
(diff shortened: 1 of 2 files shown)


lib/smtlib_lexer.mll
```diff
@@ -5,7 +5,7 @@ open Smtlib_parser
 exception SyntaxError of string
 
 let keywords =
-  let tbl = Hashtbl.create 10 in
+  let tbl = Hashtbl.create 16 in
```
zapashcanon, line 8: If you have a look at the stdlib code from OCaml, you'll see that it always takes the next power of two so this saves some time.

filipeom: Did not know that, thanks you so much! :smiley:

filipeom:
Hi @zapashcanon!

> I used some syntax magic provided by menhir. :)

Wow! It looks so much simpler and clean! Thanks for the massive improvement!  

> Also, any reason why you decided not to re-use [dolmen](https://github.com/gbury/dolmen) ?

No particular reason—I'm taking a PhD course here at IST where I proposed implementing hash-consing and other improvements to the encoding. I needed to parse some SMT2 queries to establish a baseline for next week, so I'm using the parser I had started to build before you told me about [dolmen](https://github.com/gbury/dolmen). I thought it would be quicker than learning dolmen's grammar; nevertheless, I still plan to integrate dolmen next year once I finish this course. 😄

zapashcanon:
Yes I like this syntax. Unfortunately it doesn't work with constant constructors (if gives them a unit argument and typecheck fails...).

filipeom:
> Yes I like this syntax. Unfortunately it doesn't work with constant constructors (if gives them a unit argument and typecheck fails...).

Oh yes I see, but If we don't want to mix the two syntaxes maybe we could create an helper operator that introduces a unit for us, for example: `let ( ! ) v () = v`. This way we could use it with constant constructors like: `<!Check_sat>` no?

zapashcanon:
I send an email to François, asking him if it could be handled by Menhir directly (not adding a unit argument when there's a constructor inside `<>`). Let's see what he says. :)

filipeom:
Hehe nice! But if it's impossible, I don't mind keeping both syntaxes and adjusting to `{Check_sat}` for consistency. I don't hate curly braces that much anymore. :stuck_out_tongue:

PR #50, opened 2024-01-15
Add more sizes

krtab:
Values may be wrong, but were needed to test further. LMK if I have to change them.

cc @zapashcanon

### DIFF
--- a/lib/ty.ml
+++ b/lib/ty.ml
@@ -190,4 +190,8 @@ let size (ty : t) : int =
   match ty with
   | Ty_bitv S32 | Ty_fp S32 -> 4
   | Ty_bitv S64 | Ty_fp S64 -> 8
-  | _ -> assert false
+  | Ty_bitv S8 | Ty_fp S8 -> 1
+  | Ty_int -> 4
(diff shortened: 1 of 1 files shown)


filipeom:
Oh yeah, totally forgot about extending this function for the I8s. Thanks!

zapashcanon:
Are the values for `bool` and `int` correct ? We put 4 but I'm not sure this is always true.

filipeom:
We don't use `Ty_int` in OWI, but supposedly, `Ty_int` models arithmetic integers, so they won't have a size like the other bit vector types. `Ty_bool` in OWI should always have 4 bytes. So, in a sense, outside of OWI, these might not be 100% correct all the time.

However, this function is only used to measure the size of an expression inside an `Extract` expression [here](https://github.com/formalsec/encoding/blob/3dd2f314abeed5ee5158da71da3441ff8de9c529/lib/expr.ml#L227C1-L227C29). If the pattern match falls outside of `Ty_bitv` or `Ty_fp`, we will have bigger problems because z3 will fail  with an expression that should never type.

PR #57, opened 2024-01-20
Add colibri2 support

hra687261:
With this PR, you'll be able to use Colibri2 instead of Z3 using the command line option `-p colibri2`
Support for optimization with ocplib_simplex (The simplex library used by both Colibri2 and Alt-Ergo) is incomplete, I'll take a better look at how to properly do it in another time (or someone else can if they're motivated)
Currently it's necessary to pin Colibri2 on the master branch in https://git.frama-c.com/pub/colibrics

### DIFF
--- a/bin/main.ml
+++ b/bin/main.ml
@@ -1,8 +1,4 @@
 open Encoding
-open Cmdliner
-module Z3_batch = Solver.Batch (Z3_mappings)
-module Z3_incremental = Solver.Incremental (Z3_mappings)
-module Interpret = Interpret.Make (Z3_batch)
 
 let get_contents = function
   | "-" -> In_channel.input_all In_channel.stdin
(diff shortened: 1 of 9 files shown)


bin/main.ml
```diff
@@ -14,27 +10,72 @@ let get_contents = function
...
+    ( (fun str ->
+        try
+          Ok
+            ( match str with
+            | "Z3" | "z3" -> Z3_prover
+            | "Colibri2" | "colibri2" -> Colibri2_prover
+            | _ ->
+              failwith (Fmt.str "Unsupported prover %s, try Z3 or Colibri2" str)
+            )
+        with Failure s -> Error (`Msg s) )
```
filipeom, line 28: Why do we throw the exception here? Wouldn't putting the `Error ('Msg (Fmt.str ...` in the failwith case be more straightforward? I don't usually like doing control-flow with exceptions unless there is a gain to be had, but let me know what you :)

lib/colibri2_mappings.ml
```diff
@@ -0,0 +1,989 @@
...
+            List.iter
+              (fun e ->
+                let n = Colibri2_core.Ground.convert d e in
+                Colibri2_core.Egraph.register d n;
+                Colibri2_theories_bool.Boolean.set_true d n )
+              es' )
+
+      let check s es =
+        add s es;
+        Scheduler.check_sat s.scheduler
```
filipeom, line 876: The way I encoded this using Z3 was for the expressions `es` to not be added to the solver. Because, IIRC Z3's `check` actually behaves like a `check-sat-assuming` with the argument `es` being the assumptions.

Do you think this will be problematic? We should probably add a label to `es` something like: `Val check : solver -> assumptions:expr list`

filipeom:
Does colibri2 not support ocaml 5.1? I think this might be problematic if we want to use it in Owi since it's >= 5.1

filipeom:
Btw, you don't need to support the optimization module. I just implemented it here to play with linear programming for an algorithms course project. 

But if I find time in the next few weeks, I'd probably like to give it a go. I'm interested in seeing the performance when compared with Z3's optimize (which was horrible for my use cases)

encoding.opam
```diff
@@ -35,3 +36,6 @@ build: [
 ]
 dev-repo: "git+https://github.com/formalsec/encoding.git"
 available: (arch = "x86_64" | arch = "arm64") & os != "win32" & arch != "x86_32"
+pin-depends: [
+  [ "colibri2.master" "git+ssh://git@git.frama-c.com/pub/colibrics.git#master"]
```
filipeom, line 40:
```suggestion
  [ "colibri2.master" "https://git.frama-c.com/pub/colibrics.git#master"]
```

encoding.opam.template
```diff
@@ -1 +1,4 @@
 available: (arch = "x86_64" | arch = "arm64") & os != "win32" & arch != "x86_32"
+pin-depends: [
+  [ "colibri2.master" "git+ssh://git@git.frama-c.com/pub/colibrics.git#master"]
```
filipeom, line 3:
```suggestion
  [ "colibri2.master" "https://git.frama-c.com/pub/colibrics.git#master"]
```

encoding.opam
```diff
@@ -35,3 +36,6 @@ build: [
 ]
 dev-repo: "git+https://github.com/formalsec/encoding.git"
 available: (arch = "x86_64" | arch = "arm64") & os != "win32" & arch != "x86_32"
+pin-depends: [
+  [ "colibri2.master" "https://git.frama-c.com/pub/colibrics.git#master"]
```
filipeom, line 40:
```suggestion
  [ "colibri2.master" "git+https://git.frama-c.com/pub/colibrics.git#master"]
```

encoding.opam
```diff
@@ -38,5 +38,5 @@ build: [
 dev-repo: "git+https://github.com/formalsec/encoding.git"
 available: (arch = "x86_64" | arch = "arm64") & os != "win32" & arch != "x86_32"
 pin-depends: [
-  [ "colibri2.5.1.0" "https://git.frama-c.com/pub/colibrics.git#5.1.0"]
+  [ "colibri2.dev" "https://git.frama-c.com/pub/colibrics.git#5.1.0"]
```
filipeom, line 41: I think it should be: `"git+https://git.frama-c.com/pub/colibrics.git#5.1.0"` I think without `git+` why he can't fetch the pin

hra687261: True, didn't notice it

hra687261: Well the version is also not right on the colibri2 repo

hra687261: Euh I didn't change it in the template -_-

filipeom: I think we're also missing "colibrilib". The CI is complaning. 

I think with `[ "colibrilib.dev" "git+https://git.frama-c.com/pub/colibrics.git#5.1.0"]` should be enough?

encoding.opam
```diff
@@ -38,5 +38,6 @@ build: [
 dev-repo: "git+https://github.com/formalsec/encoding.git"
 available: (arch = "x86_64" | arch = "arm64") & os != "win32" & arch != "x86_32"
 pin-depends: [
-  [ "colibri2.dev" "git+https://git.frama-c.com/pub/colibrics.git#5.1.0"]
+  [ "colibri2.dev" "git+ssh://git@git.frama-c.com/pub/colibrics.git#5.1.0"]
```
filipeom, line 41: I don't think it will work it the `ssh` link, the CI says it cannot fetch it: 

```
could not synchronize /home/runner/work/encoding/encoding/_opam/.opam-switch/sources/colibri2 from "git+ssh://git@git.frama-c.com/pub/colibrics.git#5.1.0":
```

hra687261: I can see that, but I don't think reviewing every commit as they arrive is very helpful, you can take over the PR if you want, or wait for me to fix it to review, if I don't manage to do it I'll let you know :)

filipeom: Oh so sorry, ofc I'll stop :)

hra687261:
@filipeom It seems that the CI can't install some of Colibri2's external dependencies on mac-os, so that's problematic :( I'll see what can be done about that

filipeom:
> @filipeom It seems that the CI can't install some of Colibri2's external dependencies on mac-os, so that's problematic :( I'll see what can be done about that
> 
> cf: [bobot/ocaml-flint#7](https://github.com/bobot/ocaml-flint/pull/7)

I also can't install it locally on arch. Not even with the patch you linked 😔

What we can do is make colibri2_mappings a optional library and declare colibri2 as a `depopts`. This way we can merge this so you don't have to rebase all the time. I can play with this, what do you think?

hra687261:
Yes, I agree, that seems like a good solution for now.

filipeom:
Got it. I'll also rebase my useless commits. Thanks a bunch for your help. I really appreciate it! And sorry again for any disruptions this afternoon. Won't happen again :)

hra687261:
No worries at all :)

PR #58, opened 2024-01-22
sort symbols when printing

zapashcanon:
to get a more stable output


filipeom:
Thanks! :smile:

zapashcanon:
Wait, don't merge this for now. Something is weird with the output. :sweat_smile:

filipeom:
Yeah, I'm not actually sure I like how floats are being displayed also :disappointed:

zapashcanon:
OK so the problem was that I used polymorphic compare and the type of of symbols being `{ sort : Ty.t; name : String.t }`, the `sort` field was compared first...

I defined proper compare functions to avoid this. (In a few places I still used a polymorphic compare but I believe they don't really matter).

filipeom:
Okok, makes sense! Merging, thanks!

PR #65, opened 2024-01-25
Improve hashconsing

krtab:
Closes: #52

### DIFF
--- a/lib/expr.mli
+++ b/lib/expr.mli
@@ -1,9 +1,10 @@
 (** Term definitions of the abstract syntax *)
-
-type t = private
+type inner = private
   { e : expr
   ; ty : Ty.t
   }
+
+and t = inner Hc.hash_consed
 
 and expr =
   | Val of Value.t
@@ -28,6 +29,12 @@ val pp_smt : Format.formatter -> t list -> unit
 val pp_list : Format.formatter -> t list -> unit
 val to_string : t -> string
 val simplify : ?extract:bool -> t -> t
+
+module H : sig
+  val clear : unit -> unit
+  val stats : unit -> Hashtbl.statistics
+  val length : unit -> int
+end
 
 module Bool : sig
   val v : bool -> t
--- a/dune-project
+++ b/dune-project
@@ -32,4 +32,5 @@
    (cmdliner (>= "1.2.0"))
    (zarith (>= "1.5"))
    (odoc :with-doc)
+   (hc (>= "0.3"))
    (bisect_ppx (and :with-test (>= "2.5.0") :dev))))
--- a/encoding.opam
+++ b/encoding.opam
@@ -17,6 +17,7 @@ depends: [
   "cmdliner" {>= "1.2.0"}
   "zarith" {>= "1.5"}
   "odoc" {with-doc}
+  "hc" {>= "0.3"}
   "bisect_ppx" {with-test & >= "2.5.0" & dev}
 ]
 build: [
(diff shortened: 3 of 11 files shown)


krtab:
Closes: #52

filipeom:
Depends on https://github.com/ocaml/opam-repository/pull/25120

lib/expr.mli
```diff
@@ -1,10 +1,11 @@
 (** Term definitions of the abstract syntax *)
-
-type t = private
+type inner = private
   { e : expr
   ; ty : Ty.t
   }
 
+and t = inner Hc.hash_consed
```
zapashcanon, line 7: Shouldn't we mark the three types here as `private` and only provide smart constructor ?

zapashcanon:
Adding one or two very small tests may be worth it. I'm thinking of one that checks that two values `x1` and `x2 created independently actually verifies `x1 == x2`, and one that check the number of entries in the hash-cons table (Hc provides a function `length` and even a `statistics` one).

filipeom:
>  I'm thinking of one that checks that two values x1 and x2 created independently actually verifies x1 == x2, and one that check the number of entries in the hash-cons table

I already have a small unit test for hash consing here [test/unit/test_hash_consing.ml](https://github.com/formalsec/encoding/blob/main/test/unit/test_hash_consing.ml) maybe we can built on this

lib/expr.mli
krtab: I think expr can be left non private because it doesn't have any invariant other than those of its subtypes.

I'm not sure I get what you mean with `smart constructors`, currently `@:` does the trick when needing to create a new hasconsed value?

filipeom: I think `t` doesn't need to be `private` because `hash_consed` is, right? As for `expr`, for now, we can leave it as is. When I refactor the types per Arthur's recommendation, I can complete the `smart constructors` at the end of the file :smiley:

filipeom: Btw, do you think `Value.t` should be hash-consed as well?

krtab: I discussed this with @lthls and it is indeed hash_consed that should be made private @zapashcanon

krtab: > Btw, do you think `Value.t` should be hash-consed as well?

Only if a) they are used outside on an expr.Val *and* b) the string variant is used often with identical string I think.

filipeom: > Only if a) they are used outside on an expr.Val _and_ b) the string variant is used often with identicals string I think.

Oh okok, It's they are only ever used for models outside of `expr` and these values don't typically live for very long so it should be fine. I'm fine with merging this :)

zapashcanon: OK, it was not private in `hc` actually haha (I wrote it a long time ago :S). It should be fixed in https://git.zapashcanon.fr/zapashcanon/hc/commit/9a39243a3d010adca94a37abb2567bef749b1c07 I'll release it soon it the change is OK for you.

filipeom: > OK, it was not private in `hc` actually haha (I wrote it a long time ago :S). It should be fixed in https://git.zapashcanon.fr/zapashcanon/hc/commit/9a39243a3d010adca94a37abb2567bef749b1c07 I'll release it soon it the change is OK for you.

I don't have any objections :smiley: It looks great to me :)

zapashcanon: Should be good once https://github.com/ocaml/opam-repository/pull/25125 is merged.

PR #72, opened 2024-02-08
Add option to not print values

krtab:
Useful to make output deterministic for CI

### DIFF
--- a/lib/model.ml
+++ b/lib/model.ml
@@ -17,12 +17,19 @@ let pp_print_hashtbl ~pp_sep pp_v fmt v =
   let l = Hashtbl.to_seq v |> List.of_seq |> List.sort compare_bindings in
   Format.pp_print_list ~pp_sep pp_v fmt l
 
-let pp_bindings fmt model =
+let pp_bindings fmt ?(no_values = false) model =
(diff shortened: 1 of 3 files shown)


lib/model.ml
```diff
@@ -17,12 +17,20 @@ let pp_print_hashtbl ~pp_sep pp_v fmt v =
...
-      Format.fprintf fmt "(%a %a)" Symbol.pp key Value.pp data )
+      if print_val then Format.fprintf fmt "(%a %a)" Symbol.pp key Value.pp data
+      else
+        let t = Symbol.type_of key in
+        Format.fprintf fmt "(%a %a)" Symbol.pp key Ty.pp t )
     fmt model
 
-let pp fmt model = Format.fprintf fmt "(model@\n  @[<v>%a@])" pp_bindings model
-let to_string (model : t) : String.t = Format.asprintf "%a" pp model
+let pp fmt ?(no_values = false) model =
```
zapashcanon, line 30: Having both `?(print_val = true)` and `?(no_values = false)` is not ideal. Wouldn't it be better to have the same name in both places ?

filipeom:
Thanks! :smiley:

PR #76, opened 2024-02-18
Canonization and SAT cache

filipeom:
More than 50% of Collections-C queries are duplicated, so merge this PR if the results of canonization + SAT cache are good


zapashcanon:
I read the code quickly and it looks good to me. :)

filipeom:
Canonisation is still to experimental. Closing in favour of #105

PR #77, opened 2024-02-21
Some fixes for num.ml

krtab:


lib/num.ml
```diff
@@ -7,23 +7,25 @@ type t =
...
   match (n1, n2) with
+  | I8 i1, I8 i2 -> i1 = i2
   | I32 i1, I32 i2 -> i1 = i2
   | I64 i1, I64 i2 -> i1 = i2
-  | F32 i1, F32 i2 -> i1 = i2
-  | F64 i1, F64 i2 -> i1 = i2
-  | _ -> false
+  | F32 i1, F32 i2 -> Int32.float_of_bits i1 = Int32.float_of_bits i2
+  | F64 i1, F64 i2 -> Int64.float_of_bits i1 = Int64.float_of_bits i2
+  | I8 _, _ | I32 _, _ | I64 _, _ | F32 _, _ | F64 _, _ -> assert false
```
filipeom, line 15: Should we actually fail with `assert false` or rather return `false`? My thinking was that `I8 _ = I32 _` should be false? 

One could even argue that if I32 or I64 with a value < 256 it could be compared with an I8? I just thought this, what do you think?

filipeom:
Also, perhaps we could add this regression test? Maybe I can add it afterwards so you don't have more work fixing my mess :sweat_smile: 

```ocaml
let () =
  let open Num in
  assert (
    let nan0 = F32 (Int32.bits_of_float Float.nan) in
    let nan1 = F32 (Int32.bits_of_float Float.nan) in
    not Num.(nan0 = nan1))

let () =
  let open Expr.Fpa in
  assert (
    let nan0 = F32.v Float.nan in
    let nan1 = F32.v Float.nan in
    not (Expr.equal nan0 nan1) )
```

lib/num.ml
zapashcanon: Yes, I was also expecting `false` here instead of `assert false` (if you want to put this type in a hashtbl or whatever you'd need this)

filipeom:
Thanks! :raised_hands:

PR #87, opened 2024-03-09
Parametric mappings

filipeom:
An effort to make it easier to write solver mappings 

Depends on #86


filipeom:
Closed in favour of #92

PR #96, opened 2024-03-22
Add missing/incomplete simplifications

joaomhmpereira:


lib/expr.ml
```diff
@@ -271,7 +271,9 @@ let rec binop ty (op : binop) (hte1 : t) (hte2 : t) : t =
     | Add | Or | Sub -> hte1
     | And | Mul -> make (Val (Num (I32 0l)))
     | _ -> make (Binop (ty, op, hte1, hte2)) )
-  | Binop (ty, op2, x, { node = Val (Num _ as v1); _ }), Val (Num _ as v2) -> (
+  | Binop (ty, op2, x, { node = Val (Num _ as v1); _ }), Val (Num _ as v2)
+  | Binop (ty, op2, x, { node = Val (Int _ as v1); _ }), Val (Int _ as v2)
+  | Binop (ty, op2, x, { node = Val (Real _ as v1); _ }), Val (Real _ as v2) -> (
```
filipeom, line 276:
```suggestion
  | Binop (ty, op2, x, { node = Val v1; _ }), Val v2 -> (
```

I think we can do it like this now

filipeom (approved):
Thanks! 

Before merging, could you add tests for this simplification case here: https://github.com/formalsec/encoding/blob/main/test/unit/test_simplify.ml.

Issue #99, opened 2024-03-25
Bring ECMA-SL's concrete evaluation to encoding


filipeom:
Needed in order to allow for the completion of formalsec/ECMA-SL#20

- :ballot_box_with_check: Exists
- :x: TODO
- :no_entry: Will not support 
- :exclamation: Might exist, needs verification
- :question: Don't know yet, don't add for now

## Unary

| Operators | Status | New Name |
|-----------|--------|----------|
| Typeof | :no_entry:  |
| Neg | :ballot_box_with_check: |
| BitwiseNot | :ballot_box_with_check: |
| LogicalNot | :ballot_box_with_check: |
| IntToFloat | :ballot_box_with_check:  |
| IntToString | :ballot_box_with_check: |
| IntToFourHex | :no_entry:  |
| OctalToDecimal | :no_entry:  |
| FloatToInt | :ballot_box_with_check:   |
| FloatToString | :ballot_box_with_check:   |
| ToInt | :no_entry:  |
| ToInt32 | :no_entry:  |
| ToUint16 | :no_entry:  |
| ToUint32 | :no_entry:  |
| IsNaN | :ballot_box_with_check: |
| StringToInt  | :ballot_box_with_check: |
| StringToFloat | :ballot_box_with_check:  |
| FromCharCode | :ballot_box_with_check: |
| FromCharCodeU | :no_entry:  |
| ToCharCode | :ballot_box_with_check: |
| ToCharCodeU | :no_entry:  |
| ToLowerCase | :no_entry:  |
| ToUpperCase | :no_entry:  |
| Trim | :ballot_box_with_check: |
| StringLen | :ballot_box_with_check: | Length |
| StringLenU | :no_entry:  |
| StringConcat | :ballot_box_with_check: |
| ObjectToList | :no_entry:  |
| ObjectFields | :no_entry:  |
| ArrayLen | :no_entry:  |
| ListToArray | :no_entry:  |
| ListHead | :x:  | Head |
| ListTail | :x:   | Tail |
| ListLen | :x:  | Length | 
| ListSort | :no_entry:   |
| ListReverse | :x:   | Reverse |
| ListRemoveLast | :no_entry:  |
| TupleFirst | :x:  | Head |
| TupleSecond | :x:  | Tail |
| TupleLen | :x: | Length |
| FloatToByte | :no_entry:  |
| Float32ToLEBytes | :no_entry:  |
| Float32ToBEBytes | :no_entry:  |
[...]

PR #101, opened 2024-03-28
Extend concrete evaluation

joaomhmpereira:
to address https://github.com/formalsec/encoding/issues/99


filipeom:
Hey João, sorry for the massive delay. However, in an effort to reduce your workload, I've updated issue #99. I believe we've completed most of the operators, but please pay special attention to the ones annotated with :exclamation:, as I'm not sure if we've included them yet.

In this PR, you can remove the float operators you added. I've begun moving them directly to ecma-sl here: [formalsec/ECMA-SL#stdlib](https://github.com/formalsec/ECMA-SL/tree/stdlib)

Regarding the string operators you included, I'd like to keep them, excluding the lowercase/uppercase functions. We'll also transfer those to ecma-sl's standard library.

lib/value.ml
```diff
@@ -43,6 +47,8 @@ let pp fmt (v : t) =
   | Real x -> fprintf fmt "%F" x
   | Num x -> Num.pp fmt x
   | Str x -> Format.fprintf fmt "%S" x
+  | List _ -> assert false
+  | Tuple _ -> assert false
```
filipeom, line 51:
```suggestion
  | List l -> 
    fprintf fmt "[%a]" 
      (pp_print_list ~pp_sep:(fun fmt () -> pp_print_string fmt ", ")
      pp)
      l
  | Tuple _ -> same but with "(%a)"
```

You would need to add `rec` to the definition as well

filipeom (commented):
By the way, could you add simple tests to `test_{unop|binop|triop|cvtop}`, similar to the ones already present, just to verify if the construction of the operators you added is being correctly simplified in practice?

I don't want to rush this. I'm going to pin this branch to https://github.com/formalsec/ECMA-SL/tree/stdlib and see if we can start the value migration

joaomhmpereira:
@filipeom I think this is finally ready for review

lib/ty.ml
```diff
@@ -124,6 +131,12 @@ type cvtop =
   | String_from_code
   | String_to_int
   | String_from_int
+  | String_to_float
+
+type naryop =
+  | AndN
+  | OrN
```
filipeom, line 138: These can only be used with logical operators so we can probably rename them to:

```suggestion
  | Logand
  | Logor
```

lib/ty.ml
```diff
@@ -72,13 +76,15 @@ type binop =
...
-  | Seq_concat
-  | Seq_prefix
-  | Seq_suffix
-  | Seq_contains
-  | Seq_last_index
+  | String_prefix
+  | String_suffix
+  | String_contains
+  | String_last_index
+  | String_split
```
filipeom, line 87:
```suggestion
```

I didn't see a String_split in `Eval`. Maybe we can remove this then?

filipeom:
Sorry for the delay, I did a quick check of the code, I'll try to give more feedback as I start to get more time

joaomhmpereira:
> There are a few more changes I want to add here before we merge. @joaomhmpereira you merged main into here on your last commit [fe136d2](https://github.com/formalsec/smtml/pull/101/commits/fe136d27c5e91545ed2b2cfbab51e49a2b1e4949)?

Yes, but it was an older version where the tests for the solver were not fixed yet

filipeom:
> Yes, but it was an older version where the tests for the solvers were not fixed yet

Okok, so I'll rebase it on main to fix conflicts as well

filipeom:
Don't know why the CI is failing on bitwuzla. But it should be ok

PR #106, opened 2024-04-15
cvc5 mappings

joaomhmpereira:
Addition of mappings for the cvc5 solver


lib/mappings_intf.ml
```diff
@@ -172,11 +172,25 @@ module type M = sig
...
+
     val at : term -> pos:term -> term t
 
     val concat : term -> term -> term t
 
+    val contains : term -> sub:term -> term t
+
+    val prefix : term -> s:term -> term t
+
+    val suffix : term -> s:term -> term t
```
filipeom, line 187:
```suggestion
    val is_prefix : term -> prefix:term -> term t

    val is_suffix : term -> suffix:term -> term t
```

Can we change it to this?

lib/mappings_intf.ml
```diff
@@ -172,11 +172,25 @@ module type M = sig
...
+
     val sub : term -> pos:term -> len:term -> term t
+
+    val index : term -> sub:term -> off:term -> term t
+
+    val replace : term -> s2:term -> s3:term -> term t
```
filipeom, line 193:
```suggestion
    val index_of : term -> sub:term -> pos:term -> term t

    val replace : term -> pattern:term -> with_:term -> term t
```

filipeom (approved):
Looks good, I'd prefer we used better names for the string functions. I left some suggestions below.

filipeom:
Thanks! I'm fine with merging this if you think it has everything you need :)

PR #109, opened 2024-04-18
Remove arch and os constraints

filipeom:


zapashcanon:
Yes, unfortunately `ocaml_intrinsics` is not available on windows, we have the same issue in Owi... Maybe using the new `ocaml_intrinsics_kernel` would work but I did not investigated further for now.

filipeom:
Yeah, it looks like it is available for windows https://github.com/janestreet/ocaml_intrinsics_kernel/blob/master/ocaml_intrinsics_kernel.opam 

But not for 32-bit architectures right? I'll give it a shot.

PR #130, opened 2024-06-05
Multi backend

filipeom:
Removed the optional libraries for solver mappings. Now, the solver mappings are compiled when the solver is installed. If the solver is not installed, it uses a stub module that fails during runtime.

cc @zapashcanon, I created a function called `Solver_dispatcher.mappings_of_solver` for our purpose. Please let me know if there's anything you'd like to change or improve. :smiley:


lib/solver_dispatcher.ml
```diff
@@ -0,0 +1,14 @@
+(* TODO: put this in some other more appropriate module? *)
+type solver_type =
+  | Z3_solver
+  | Z3_solver2
+  | Cvc5_solver
+  | Colibri2_solver
+  | Bitwuzla_solver
+
```
zapashcanon, line 8: What about providing a `val solver_type_of_string : string -> (solver_type, string) Result.t` along the way ?

Something like:

```ocaml
let solver_type_of_string s =
  match String.map Char.lowercase_ascii s with
  | "z3" -> Ok Z3_solver
  | "z32" -> Ok Z3_solver2 (* I don't like this name TBH, make we could make the intent more clear ? *)
  | "colibri2" -> Ok Colibri2
  | "bitwuzla" -> Ok Bitwuzla_solver
  | "cvc5" -> Ok Cvc5_solver
  | s -> Format.ksprintf Result.error "unknown solver %s" s
```

filipeom: Seems like a good idea to me. 

>  | "z32" -> Ok Z3_solver2 (* I don't like this name TBH, make we could make the intent more clear ? *)

I'll just remove `Z3_solver2` for the moment. It's to experimental, and probably buggy too

zapashcanon:
This looks great, that's exactly what I wanted, thanks!

I have one suggestion to reduce the maintenance burden. You could have a single `nop.ml` and use dune rules to generate all the `$solver.nop.ml` by copying `nop.ml` and replacing the name of the solver in the `let make_cont ()` function ?

filipeom:
> This looks great, that's exactly what I wanted, thanks!
> 
> I have one suggestion to reduce the maintenance burden. You could have a single `nop.ml` and use dune rules to generate all the `$solver.nop.ml` by copying `nop.ml` and replacing the name of the solver in the `let make_cont ()` function ?

Thanks! I pushed an attempt using `sed`. Do you know if there is a more idiomatic way of doing this with dune?

zapashcanon:
Instead of using `>` you can probably use `with-stdout-to` from dune. But I don't know about how to avoid sed. One way could be to do something like:

`echo 'let solver_name = "colibri2"\n' > colibri2.ml; cat nop.ml > colibri2.ml`

And use `solver_name` in `let make_cont`.

filipeom:
> Instead of using `>` you can probably use `with-stdout-to` from dune. But I don't know about how to avoid sed. One way could be to do something like:
> 
> `echo 'let solver_name = "colibri2"\n' > colibri2.ml; cat nop.ml > colibri2.ml`
> 
> And use `solver_name` in `let make_cont`.

I tried this and it worked. However, I discovered this approach: 

```ocaml
(rule
 (targets
  colibri2_mappings.nop.ml
  bitwuzla_mappings.nop.ml
  cvc5_mappings.nop.ml)
 (action
  (progn
   (with-stdout-to
    colibri2_mappings.nop.ml
    (echo
     "include (val (failwith \"Colibbri2 not installed!\") : Mappings_intf.S)"))
   (with-stdout-to
    bitwuzla_mappings.nop.ml
    (echo
     "include (val (failwith \"Bitwuzla not installed!\") : Mappings_intf.S)"))
   (with-stdout-to
    cvc5_mappings.nop.ml
    (echo
     "include (val (failwith \"cvc5 not installed!\") : Mappings_intf.S)")))))
```

Due to the `_intf` trick you told be about a while ago.

filipeom:
> Instead of using `>` you can probably use `with-stdout-to` from dune. But I don't know about how to avoid sed. One way could be to do something like:
> 
> `echo 'let solver_name = "colibri2"\n' > colibri2.ml; cat nop.ml > colibri2.ml`
> 
> And use `solver_name` in `let make_cont`.

This ended being the best solution so far :smiley:

zapashcanon:
I would say we could merge this! Do you think you could do a new release?

filipeom:
Will do :smile:

PR #152, opened 2024-06-17
Adds `is_available` flag to know which solvers are installed

### LINKED ISSUE #149: add `is_available` a boolean to know if we're using the `nop` version or the `default` version
For instance, it would allow to do something like:

```ocaml
let solver =
  if Z3_mappings.is_available then Ok Z3_mappings
  else if Colibri2_mappings.is_available then Ok Colibri2
  else if ...
  else Error (`Msg "no available solver")
```

It would be useful for instance to try many solvers when the user didn't specified one in Owi - instead of only trying Z3 and to fail if it's not there.

The implementation is quite straightforward (add `let is_available = false` to the `nop` file and `let is_available = true` in all `default` implementations).

filipeom:
Closes #149 

Is this enough @zapashcanon?

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,8 @@
 
 ### Added
 
+- Adds `Solver_dispacher.{is_available|available_solvers|solver}` to check
+availability of installed solvers
 - Model generation for Bitwuzla
 
 ### Changed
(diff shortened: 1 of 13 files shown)


zapashcanon:
Yes thanks a lot!

> (** Returns first available solver or errors when none exist *)

This one is probably going to be quite useful ! :)

PR #154, opened 2024-06-17
Fix argument type in cvc5 mappings

joaomhmpereira:


filipeom:
I will try to add `cvc5` workflow so that we might try to catch these errors here as well.

PR #157, opened 2024-06-21
Don't hc bool values unecessarily

filipeom:


lib/expr.ml
```diff
@@ -345,12 +345,10 @@ let triop' (ty : Ty.t) (op : triop) (e1 : t) (e2 : t) (e3 : t) : t =
...
-  match (view e1, view e2, view e3) with
-  | Val v1, Val v2, Val v3 -> value (Eval.triop ty op v1 v2 v3)
-  | Val v, _, _ -> (
-    match op with
-    | Ite -> ( match v with True -> e2 | False -> e3 | _ -> assert false )
-    | _ -> triop' ty op e1 e2 e3 )
+  match (op, view e1, view e2, view e3) with
+  | op, Val v1, Val v2, Val v3 -> value (Eval.triop ty op v1 v2 v3)
+  | Ite, Val True, _, _ -> e2
+  | Ite, Val False, _, _ -> e3
```
zapashcanon, line 351: Should'nt they be in this order:

```ocaml
  | Ite, Val True, _, _ -> e2
  | Ite, Val False, _, _ -> e3
  | op, Val v1, Val v2, Val v3 -> value (Eval.triop ty op v1 v2 v3)
```

Also you can merge the two cases like this:

```ocaml
  | (Ite, Val True, e, _) | (Ite, Val False, _, e) -> e
  | op, Val v1, Val v2, Val v3 -> value (Eval.triop ty op v1 v2 v3)
```

lib/expr.ml
```diff
@@ -515,41 +513,45 @@ module Bool = struct
...
+
+  let false_ = value False
+
+  let to_val b = if b then true_ else false_
 
-  let not (b : t) =
+  let not_ (b : t) =
     match of_val (view b) with
-    | Some b -> v (not b)
+    | Some b -> to_val b
```
zapashcanon, line 524: `to_val (not b)` ?

lib/expr.ml
```diff
@@ -515,41 +513,45 @@ module Bool = struct
...
-    | _ -> relop' Ty_bool Ne b1 b2
+    | Val True, Val False | Val False, Val True -> true_
+    | _ -> relop Ty_bool Ne b1 b2
 
   let and_ (b1 : t) (b2 : t) =
     match (of_val (view b1), of_val (view b2)) with
-    | Some b1, Some b2 -> v (b1 && b2)
+    | Some b1, Some b2 -> to_val (b1 && b2)
     | Some true, _ -> b2
     | _, Some true -> b1
```
zapashcanon, line 544: You can merge the two cases:

`| Some true, b | b, Some true -> b`

lib/expr.ml
```diff
@@ -515,41 +513,45 @@ module Bool = struct
...
-    | Some false, _ | _, Some false -> value False
-    | _ -> binop' Ty_bool And b1 b2
+    | Some false, _ | _, Some false -> false_
```
zapashcanon, line 545: I believe the order should also be:

```ocaml
    | Some true, _ -> b2
    | _, Some true -> b1
    | Some false, _ | _, Some false -> false_
    | Some b1, Some b2 -> to_val (b1 && b2)
```

lib/expr.ml
```diff
@@ -515,41 +513,45 @@ module Bool = struct
...
 
   let or_ (b1 : t) (b2 : t) =
     match (of_val (view b1), of_val (view b2)) with
-    | Some b1, Some b2 -> v (b1 || b2)
+    | Some b1, Some b2 -> to_val (b1 || b2)
     | Some false, _ -> b2
     | _, Some false -> b1
-    | Some true, _ | _, Some true -> value True
-    | _ -> binop' Ty_bool Or b1 b2
+    | Some true, _ | _, Some true -> true_
```
zapashcanon, line 553: Same here, the order does not look good to me

lib/expr.ml
filipeom: Correct I must have missed this :swe

lib/expr.ml
filipeom: > Should'nt they be in this order:
> 
> ```ocaml
>   | Ite, Val True, _, _ -> e2
>   | Ite, Val False, _, _ -> e3
>   | op, Val v1, Val v2, Val v3 -> value (Eval.triop ty op v1 v2 v3)
> ```
> Also you can merge the two cases like this:
> 
> ```ocaml
>   | Ite, Val True, e, _ | Ite, Val False, _, e -> e
>   | op, Val v1, Val v2, Val v3 -> value (Eval.triop ty op v1 v2 v3)
> ```

Indeed, I will fix it :smile: Thanks again!

lib/expr.ml
filipeom: Does the order matter that much here? I don't see `| Some b1, Some b2 -> to_val (b1 && b2)` ever being reachable at the end?

lib/expr.ml
filipeom: Same question here :sweat_smile:

lib/expr.ml
filipeom: Unfortunately, I cannot merge this. `b` would have type `bool option` instead of `Expr.t`

lib/expr.ml
zapashcanon: Probably not, I guess the compiler should be able to tell you :)

PR #165, opened 2024-06-27
Add conflicts on solvers versions outside depopt range

krtab:
We want to nudge our users to install versions of the solvers that are actually compatible with smtml. To this end, I added a conflict with solver versions that are outside of the depopt range.


filipeom (approved):
LGTM! Thanks :smiley:

PR #175, opened 2024-07-04
Handle Nan and Infinity floats

joaomhmpereira:
Add simplification cases to handle operations with `Nan` or `Infinity` floats


filipeom:
Thanks!

PR #182, opened 2024-07-18
use prelude

zapashcanon:
It helped discovering a few bugs in Owi and I think it would be beneficial for smtml too, in particular regarding use of polymorphic operations such as equality or comparisons.

This is not working yet but I'm not far.

Once this is done, I could either remove the dependency on prelude but simply keep the modified code, or keep prelude as a dependency, as you prefer.


filipeom:
I was actually testing this out :sweat_smile:

 I would prefer keeping the dependency on prelude.

zapashcanon:
I managed to fix almost two issues. There is two left things to do:

- check some `TODO` regarding the handling of NaN, in some places I used `Float.compare` and `Float.equal` which treats NaN as equal to itself whereas before the code was using `=` of other polymorphic comparisons that are treating NaN differently;
- I did not fixed all errors in `eval.ml` when I started realizing that this was mostly the code from Owi in `primitives/*.ml` which already gave me a headache when fixing it in Owi... Should we try to think about a way to properly share this code ? Maybe releasing it into a separate library ?

I don't have the time or energy to fix these two issues right now so feel free to take over (but I can publish the code in `owi/primitives/*.ml` as a separate library if you're interested and believe it would fix the issue).

filipeom:
> I don't have the time or energy to fix these two issues right now so feel free to take over (but I can publish the code in owi/primitives/*.ml as a separate library if you're interested and believe it would fix the issue).


I can take care of the first issue :smile:. As for the second issue, if you want to release it as a separate library, I will import it here. But if you don't have the time, I can just fix it here in the same way you did for owi. Eventually, for bitvectors, I'd like to use something like [bitv](https://github.com/backtracking/bitv) or [bitvec](https://opam.ocaml.org/packages/bitvec/) so that we can do proper modulo arithmetic over arbitrarily-sized bitvecs.

zapashcanon:
Should be good with https://github.com/ocaml/opam-repository/pull/26273

filipeom:
Thanks!

PR #209, opened 2024-09-11
Add remaining parsing for Bitvectors and Floating-point arithmetic

### LINKED ISSUE #208: Parse remaining smtlib theories
- String
- Bitvector
- Floating-point

filipeom:
As well as perform some fixes to allow parsing of smtlib benchmarks.


filipeom:
Not finished, but we're already running collections-c so I'm merging has it is becoming to big.

PR #213, opened 2024-09-14
Use mdx in doc and fix unresolved refs

filipeom:


doc/index.mld
```diff
@@ -15,9 +15,45 @@ for Z3 encoding. The simplest approach to creating a solver in the Smtml
 is outlined below:
 
 {@ocaml[
-module Z3 = Smtml.Solver.Z3_incremental
-
-let solver : Z3.t = Z3.create ()
+# #use "topfind"
+- : unit = ()
+Findlib has been successfully loaded. Additional directives:
```
zapashcanon, line 20: @filipeom, you can hide a bunch of lines by having a line containing:

`...`

It allows to hide this kind of stuff you probably don't want to print. :)

filipeom: Ah perfect! Thanks :heart:

PR #214, opened 2024-09-16
Fix Colibri2 mapping of cvtop operations

hra687261:


filipeom:
Thanks very much! :relaxed:

PR #217, opened 2024-09-18
Add SMTLIB parsing for FPs

joaomhmpereira:


src/mappings.ml
```diff
@@ -120,6 +120,14 @@ module Make (M_with_make : M_with_make) : S_with_fresh = struct
           Fmt.failwith {|Bool: Unsupported Z3 relop operator "%a"|} Ty.pp_relop
             op
 
+      (* let naryop op l =
```
filipeom, line 123: Why did we comment this out?

filipeom:
Should be good to go. Letting the CI run and then I'm merging

PR #219, opened 2024-09-28
Fix parsing for .smt2 files

joaomhmpereira:


bin/main.ml
```diff
@@ -58,6 +74,23 @@ let test debug solver prover_mode print_statistics files =
     match result with Error (`Msg e) -> failwith e | Ok state -> state
   and test_files files = List.fold_left test_path None files in
   let state = test_files files in
+  let write_exception_log () =
+    let oc = open_out "exceptions.log" in
```
filipeom, line 78:
```suggestion
    Out_channel.with_open_text "exceptions.log" @@ fun oc ->
```

And you can remove the `close_out` at the end

PR #221, opened 2024-10-03
Caches consts

filipeom:
Adds `caches_consts` flag to mappings
    
    Solvers can inform us that they cache constants by setting the flag
    `caches_consts = true` in their mappings. This should avoid having a
    `Map.find_opt` everytime we call `make_symbol`.
    
    Currently the mappings that behave like this are only Z3

### DIFF
--- a/src/ast/ty.mli
+++ b/src/ast/ty.mli
@@ -178,7 +178,7 @@ val naryop_equal : naryop -> naryop -> bool
 
 val equal : t -> t -> bool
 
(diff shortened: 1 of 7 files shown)


filipeom:
@joaomhmpereira I've set cvc5's `caches_const=true` so this should solve some performance issues?

filipeom:
Queueing to merge. We can set cvc5's caches_const to true in another PR

PR #226, opened 2024-10-07
Update benchmarking dataset submodules

filipeom:
Adds two directories to `bench/` for submodules:

- https://github.com/formalsec/smt-collections-c
- https://github.com/formalsec/smt-comp


filipeom:
@joaomhmpereira  Is this correct? Am I missing some submodule?

joaomhmpereira:
Yes, I think it's correct

PR #249, opened 2024-11-18
Update benchpress pin

filipeom:


filipeom:
Macos CI is broken due to `pkgconf` (see https://github.com/ocaml/setup-ocaml/issues/897)

PR #258, opened 2024-12-03
fix bug in parsing of Num.t

zapashcanon:
spotted while writing some tests for `owi replay` (I'm still not able to run the tests for smtml so, if the CI breaks I'll need your help again, sorry haha...)


filipeom:
Seems I can only write bugs :sweat_smile: 

The CI should be good, I don't have tests for this function atm. Thanks!

zapashcanon:
Yes don't worry, now it'll be tested by Owi :-)

PR #270, opened 2025-01-24
update for future prelude and scfg

zapashcanon:
Hi ! :)

I made some changes to both libraries:

- in `scfg`, I'm now using ``` Error (`Msg "something")``` instead of `Error "something"`;
- in `prelude`, I'm decided to change the design of the library quite a bit so there's a lot of change, the idea is to be able to use the library only in development mode and not to depend on it in release mode, that is, we should only *hide* stuff from the stdlib, and not add new functions or rename existing ones ; it's not yet fully done but most of it is, I also opened a few PR against ocaml/ocaml in order to add stuff I think is missing.

They're not yet on opam but I wanted to test them here and in Owi. I'm making the PR now so I don't lost the changes and you can have a first look.

There is a few places where I left a comment for you (with a `TODO`). In particular, there's something I don't understand about a stack somewhere.

Cheers!


filipeom:
Hi! :smiley: 

> in prelude, I'm decided to change the design of the library quite a bit so there's a lot of change, the idea is to be able to use the library only in development mode and not to depend on it in release mode, that is, we should only hide stuff from the stdlib, and not add new functions or rename existing ones ; it's not yet fully done but most of it is, I also opened a few PR against ocaml/ocaml in order to add stuff I think is missing.

I think this is a great design decision going forward! Thank you so much for the PR!! 

> There is a few places where I left a comment for you (with a TODO). In particular, there's something I don't understand about a stack somewhere.

Okok, I'll try to respond to some in the review.

Thanks again!

Cheers!

src/eval.ml
```diff
@@ -370,20 +378,26 @@ module Lst = struct
...
     let op' = `Binop op in
     match op with
-    | At -> (
+    | At ->
       let lst = of_value 1 op' v1 in
       let i = Int.of_value 2 op' v2 in
-      try List.nth lst i
-      with Failure _ | Invalid_argument _ -> raise IndexOutOfBounds )
+      begin
+        (* TODO: change datastructure? *)
```
filipeom, line 396: :+1: I'll create an issue for this

src/mappings.ml
```diff
@@ -745,13 +749,17 @@ module Make (M_with_make : M_with_make) : S_with_fresh = struct
...
       let push { solver; ctx; _ } =
-        let top = Stack.top ctx in
-        Stack.push top ctx;
-        M.Solver.push solver
+        (* TODO: I don't understand why we are duplicating the topmost element of ctx, is this actually the expected behavior? *)
+        match Stack.top_opt ctx with
+        | None -> assert false
+        | Some top ->
+          Stack.push top ctx;
+          M.Solver.push solver
```
filipeom, line 757: > (* TODO: I don't understand why we are duplicating the topmost element of ctx, is this actually the expected behavior? *)

Yes, since our context is immutable, when we perform a `push` we duplicated the context so when we later backtrack (with a `pop`) to the context before the push

filipeom:
This is good to go?

zapashcanon:
Not yet! I still have a few things to change :) I'll let you know once I'm done

PR #288, opened 2025-02-24
Add `Expr.Set.hash` function as an alias to `Expr.Set.to_int`

filipeom:


zapashcanon:
Thanks for taking the time do to this one! :D

filipeom:
> Thanks for taking the time do to this one! :D

Hehe, np it was an easy one to do :sweat_smile:

PR #291, opened 2025-02-27
Add extract simplifications

krtab:
I'm not sure its working yet but opening it to discuss it


zapashcanon:
@filipeom I'm not sure if this is still useful?

filipeom:
It could be, I'd have to go back and check the expressions this was trying to simply :grimacing:

PR #306, opened 2025-03-04
Add `get_sat_model` and improve `model` documentation

### LINKED ISSUE #304: add `check` into `model`
A typical code asking for a model will look like:

```ocaml
match Solver.check_set s pc with
| `Sat -> begin
  match Solver.model ?symbols s with
  | None -> assert false
  | Some model -> model
  end
| `Unsat -> assert false
| `Unknown -> assert false
```

Because it is actually *required* to call `check` or `check_set` before calling `model`. (Can you confirm this is right?)

I would like not to have to think about this for two reasons: it is easy to forget, and it is easy to remove the call by doing some kind of optimization (for instance, memoizing the call to check :-)).

Would you be OK about either adding a call to `check[_set]` to `model` or adding a new function that does so?

filipeom:
- Add comprehensive documentation for new `get_sat_model` function
- Improve clarity and structure of `model` documentation
- Add cross-references between `model` and `get_sat_model`
- Clarify usage for cached vs. non-cached solvers

Closes #304

### DIFF
--- a/src/smtml/solver.ml
+++ b/src/smtml/solver.ml
@@ -55,6 +55,16 @@ module Base (M : Mappings_intf.S) = struct
   let model ?(symbols : Symbol.t list option) (s : M.solver) : Model.t option =
     let+ model = M.Solver.model s in
     M.values_of_model ?symbols model
+
+  let get_sat_model ?symbols s set =
+    match check_set s set with
+    | `Sat -> (
+      match model ?symbols s with
+      | Some _ as model -> model
+      | None ->
+        (* Should never happen *)
+        assert false )
+    | `Unsat | `Unknown -> None
 end
 
 module Incremental (M : Mappings_intf.S) : Solver_intf.S =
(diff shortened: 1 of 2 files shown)


src/smtml/solver.ml
```diff
@@ -181,8 +197,21 @@ module Cached (Mappings_ : Mappings.S) = struct
...
     let get_statistics (s : t) : Statistics.t = get_statistics s.solver
 
+    let get_sat_model ?symbols s set =
+      let assert_ = Expr.Set.union set s.top in
+      get_sat_model ?symbols s.solver assert_
+
+    let model ?(symbols : Symbol.t list option) (s : t) : Model.t option =
+      let open Option in
+      let* last_check = s.last_check in
+      (* We need to explicitly check_set because we don't want a cached reseponse *)
```
zapashcanon, line 207: s/reseponse/response/

filipeom:
```suggestion
      (* We need to explicitly check_set because we don't want a cached response *)
```

PR #323, opened 2025-04-18
parse booleans

S41d:
This PR fixes a problem I had where `Smtml.Model.Parse.Scfg.from_file` couldn't parse models that contained boolean values like this one :
```scfg
model {
  symbol symbol_0 i1 false
}
```


filipeom:
Ah, it turns out the hack I did here:

https://github.com/formalsec/smtml/blob/008b65d24e3e0f564ab73775441b851454cefd14/src/smtml/mappings.ml#L719-L724

Is causing more unpleasant consequences. Hopefully, this won't be necessary once I have time to finish #284.

Thanks for the fix!

PR #326, opened 2025-04-25
changing back so smtml can use Alt-Ergo again

felixL-K:


dune-project
```diff
@@ -46,6 +46,8 @@
  (conflicts
   (bitwuzla-cxx
    (< "0.6.0"))
+  (alt-ergo-lib
+   (>= "2.6.1"))
```
zapashcanon, line 50: It should be "< 2.6.1" (we are adding a conflict on all unsupported version rather than saying which version are valid in depopts above)

filipeom:
Thanks! :smiley:

Issue #163, opened 2024-06-25
No mappings for `to_ieee_bv` operator in Bitwuzla


filipeom:
Possible solution from smt-lib:

```smt2
:note
 "There is no function for converting from (_ FloatingPoint eb sb) to the
  corresponding IEEE 754-2008 binary format, as a bit vector (_ BitVec m) with 
  m = eb + sb, because (_ NaN eb sb) has multiple, well-defined representations.
  Instead, an encoding of the kind below is recommended, where f is a term
  of sort (_ FloatingPoint eb sb):

   (declare-fun b () (_ BitVec m))
   (assert (= ((_ to_fp eb sb) b) f))
 "
```

zapashcanon:
cc @felixL-K, do you want to try implementing it in Smt.ml?

PR #331, opened 2025-05-02
implement function to_ieee_bv for Bitwuzla

felixL-K:
I’m working on the to_ieee_bv function to convert a floating-point number to its IEEE-754 bit vector form for Bitwuzla.

Right now, I'm trying to figure out whether I should create a new mk_bv_value_float function in the kind module or if it’s better to just use Bitwuzla’s existing API for this.

I'd love some feedback on:
Should I add mk_bv_value_float, or can I achieve this with the current API?
Any advice on how to handle floating-point to bit vector conversion using Bitwuzla’s tools?
If you have any thoughts on how to improve the implementation, feel free to share!

Thanks for your help!

### DIFF
--- a/src/smtml/bitwuzla_mappings.default.ml
+++ b/src/smtml/bitwuzla_mappings.default.ml
@@ -407,8 +407,29 @@ module Fresh_bitwuzla (B : Bitwuzla_cxx.S) : M = struct
     let of_ieee_bv eb sb bv = mk_term1_indexed2 Kind.Fp_to_fp_from_bv bv eb sb
 
     (* TODO *)
-    let to_ieee_bv _ =
-      Fmt.failwith "Bitwuzla_mappings: to_ieee_bv not implemented"
+    let to_ieee_bv x (* x : Term.IEEE_754 -> Term.Z *) =
+      let (sign,exp,mant) = match Term.kind x with
+        | Kind.Fp_fp ->
+           begin match x with
+           | Term.IEEE_754 (x,y,z) -> (x,y,z)
+           | _ -> Fmt.failwith "Bitwuzla_mappings: to_ieee_bv problem"
+           end
+        | _ -> Fmt.failwith "Bitwuzla_mappings: to_ieee_bv problem"
+      in
+      let v = (Float.pow 2. exp) *. mant *. (Float.pow (-1.) sign) in
+      mk_bv_value_int (mk_bv_sort 64) (Float.to_int v)
+
+    (*     begin match Term.children x with *)
+    (*       | [|x;y;z|] -> (x,y,z) *)
+    (*       | _ -> Fmt.failwith "Bitwuzla_mappings: to_ieee_bv problem" *)
+    (*     end *)
+    (* | _ -> Fmt.failwith "Bitwuzla_mappings: to_ieee_bv problem" *)
+    (* in *)
+    (* let xx = Term.value x in *)
+    (* let v = (Float.pow 2. exp) *. mant *. (Float.pow (-1.) sign) in *)
+    (* mk_bv_value_int (mk_bv_sort 64) (Float.to_int v) *)
+
+    (* Fmt.failwith "Bitwuzla_mappings: to_ieee_bv not implemented" *)
   end
 
   module Func = struct
--- a/src/smtml/bitwuzla_mappings.default.ml
+++ b/src/smtml/bitwuzla_mappings.default.ml
@@ -407,29 +407,35 @@ module Fresh_bitwuzla (B : Bitwuzla_cxx.S) : M = struct
     let of_ieee_bv eb sb bv = mk_term1_indexed2 Kind.Fp_to_fp_from_bv bv eb sb
 
     (* TODO *)
-    let to_ieee_bv x (* x : Term.IEEE_754 -> Term.Z *) =
-      let (sign,exp,mant) = match Term.kind x with
-        | Kind.Fp_fp ->
-           begin match x with
-           | Term.IEEE_754 (x,y,z) -> (x,y,z)
-           | _ -> Fmt.failwith "Bitwuzla_mappings: to_ieee_bv problem"
-           end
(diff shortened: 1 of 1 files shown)


zapashcanon:
Fix #163

src/smtml/bitwuzla_mappings.default.ml
```diff
@@ -407,8 +407,35 @@ module Fresh_bitwuzla (B : Bitwuzla_cxx.S) : M = struct
...
 
     (* TODO *)
-    let to_ieee_bv _ =
-      Fmt.failwith "Bitwuzla_mappings: to_ieee_bv not implemented"
+    let to_ieee_bv x =
+      let (sign, exp, mant) = Term.value Term.IEEE_754 x in
+      let bits_str = String.concat [sign; exp; mant] in
+      let bv = Z.of_string_base 2 bits_str in
+      let total_bits = String.length bits_str in
+      mk_bv_value_int (mk_bv_sort total_bits) (Float.to_int bv)
```
zapashcanon, line 415: The CI is saying that `Float.to_int bv` is incorrect because `bv` has type `Z.t` and not `float`. Why not using `Z.to_int` for instance? (If you need 64 bits you can use `Z.to_int64` but then maybe you need a `mk_bv_value_int64` or something?)

filipeom:
Thanks for the PR! 

I think going in this direction isn't going to solve our problem, you're using `Term.value` which it will try to get a term representing the model value of the term `x`. This only works, if we already called `check_sat`. And also, we would be concretizing here, which we want to avoid doing.

I think a simpler solution to start with is to create an uninterpreted function which simply does: `float32 -> i32`. And the `to_ieee_bv` does an application of such function to `x`. 

To do this, I would first define the uninterpreted function symbol:

```ocaml
let f32_to_i32 =
   let args = [| Types.float 8 24 |] in
   let ret = Types.bitv 32 in
   let sort = B.mk_fun_sort args ret in
   B.mk_const sort ~symbol:"f32_to_i32"
```

Then, define the to_ieee_bv function as an application of this symbol:

```ocaml
let to_ieee_bv f = mk_term2 Kind.Apply f32_to_i32 f
```

You can experiment with this by writing a test for the Bitwuzla solver like this:

```ocaml
let solver =
   Solver.create ~params:(Params.default ()) ~logic:Logic.QF_UFBV ()
in
let x = symbol "x" (Ty_fp 32) in
let y = symbol "y" (Ty_bitv 32) in
let converted = Expr.cvtop (Ty_bitv 32) Reinterpret_float x in
Solver.add solver [ Expr.relop Ty_bool Eq converted y ];
assert_sat ~f:"test_i32_of_f32" (Solver.check solver [])
```

Alternatively, you can experiment with uninterpreted functions directly in SMT-LIB:

```smt2
; test.smt2
(set-logic QF_UFBV)

(declare-fun f ((_ BitVec 32)) (_ BitVec 32))
(declare-const x (_ BitVec 32))

(assert (= (f x) #x0000002a))

(check-sat)
(get-model)
```

Run it in Z3 to get a model:

```sh
$ z3 test.smt2
sat
(
  (define-fun x () (_ BitVec 32)
    #x00000000)
  (define-fun f ((x!0 (_ BitVec 32))) (_ BitVec 32)
    #x0000002a)
)
```

felixL-K:
Thank you very much for your feedback and explanations!

I just don't understand why your example that works with z3, doesn't work with smtml  : 
```
$ smtml run test.smt2 
smtml: internal error, uncaught exception:
       Uncaught exception:
       Failure("41-95: fun_decl f () ((FloatingPoint 8 24)) (BitVec 32)")
       with backtrace:
       
       
```

filipeom:
> I just don't understand why your example that works with z3, doesn't work with smtml :

It's to do with our frontend, currently the parser only supports parsing constants (i.e., `fun_decl`s without arguments):

https://github.com/formalsec/smtml/blob/2cb771540f04813000eda32a53b4244ee7565528/src/smtml/smtlib.ml#L339-L344

So it's falling into the pattern match where it has arguments and raises the assertion. It's in our plans to support these declarations, but we rarely use them so that's why they're not working.

src/smtml/bitwuzla_mappings.default.ml
```diff
@@ -407,8 +407,13 @@ module Fresh_bitwuzla (B : Bitwuzla_cxx.S) : M = struct
...
     (* TODO *)
-    let to_ieee_bv _ =
-      Fmt.failwith "Bitwuzla_mappings: to_ieee_bv not implemented"
+    let f32_to_i32 =
+      let args = [| Types.float 8 24 |] in
+      let ret = Types.bitv 32 in
+      let sort = B.mk_fun_sort args ret in
+      B.mk_const sort ~symbol:"f32_to_i32"
+
+    let to_ieee_bv f = mk_term2 Kind.Apply f32_to_i32 f
```
filipeom, line 416:
```suggestion
    let to_ieee_bv f =
      let f_size = fp_exp_size f + fp_sig_size f in
      if f_size = 32 then
        mk_term2 Kind.Apply f32_to_i32 f
      else if f_size = 64 then 
        mk_term2 Kind.Apply f64_to_i64 f
      else assert false (* Or raise a failure with a nice error message *)
```

Maybe you could do something similar to this to extend it to support both f32 and f64 :smile:

filipeom:
Thanks! I want to compare this with an alternative encoding of `to_ieee_bv` using an `ite` expression. I'll run the failing benchmarks of test-comp with this encoding and the alternative and then I'll give you feedback on it :smiley:

zapashcanon:
IIRC this function is also missing in alt-ergo and/or CVC5. Do you prefer to share the implementation for the three solver? If yes, maybe you should also have a look at the other solvers while looking how the current and the `ite` version perform on the failing Test-Comp tests?

filipeom:
I'm happy to merge this for the time being, but we should create an issue to track it, as it might produce false positives with Bitwuzla, raising failures when none would otherwise occur. That said, it will still reliably raise a failure when one does exist.

filipeom (approved):

PR #332, opened 2025-05-02
Experiment with safe expressions

filipeom:
cc @zapashcanon 

I think this is harder then I thought, just including `expr.ml` and making `type expr = private` is not enough, now the solver needs to be parametric on an expression module so that it can take both `Safe_expr` and `Expr`.

Any ideas that wouldn't lead to code duplication?

### DIFF
--- a/src/smtml/expr.ml
+++ b/src/smtml/expr.ml
@@ -308,6 +308,8 @@ let value (v : Value.t) : t = make (Val v) [@@inline]
 
 let ptr base offset = make (Ptr { base; offset })
 
+let list l = make (List l)
+
 let app symbol args = make (App (symbol, args))
 
 let[@inline] binder bt vars expr = make (Binder (bt, vars, expr))
--- a/src/smtml/expr.ml
+++ b/src/smtml/expr.ml
@@ -320,11 +320,11 @@ let forall vars body = binder Forall vars body
 
 let exists vars body = binder Exists vars body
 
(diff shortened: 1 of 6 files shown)


zapashcanon:
I see two possibilities.

In `safe_expr.mli`, maybe something like:

```ocaml
type t = Expr.t = expr Hc.hash_consed
```

Is enough to let us pass either an `Expr.t` or a `Safe_expr.t` to the various solver's functions?

The other one is to only have `Safe_expr` and not `Expr`?

filipeom:
> The other one is to only have `Safe_expr` and not `Expr`? (or rather, simply make `Expr` safe by making its types private directly)

I think this might be better no? I think I'll move forward with this then :smiley:

zapashcanon:
Great, and I like the new naming convention, with `raw` for the unsimplified case. :)

filipeom:
Yes, it's better this way. I found a lot of places in my other tool where I wasn't simplifying when I should :sweat_smile: Thanks, this was an awesome suggestion!

> Great, and I like the new naming convention, with `raw` for the unsimplified case. :)

I used `raw_` as it was the shortest prefix I could think of :sweat_smile: But I also considered `plain_`, `basic_`, and `direct_`. If you like another one that is better then `raw_` I can change it :smiley:

zapashcanon:
No, I believe `raw` is great. I tried with this PR in Owi, and there's also a nice side-effect, being that forcing the use of the smart constructors makes the code more readable: https://github.com/OCamlPro/owi/pull/678

PR #335, opened 2025-05-05
Use mappings.ml for Colibri2

hra687261:
I made this change so that I don't have to copy definitions of unsupported symbols.
I will do the same with Alt-Ergo once #326 is merged

### DIFF
--- a/src/smtml/dolmenexpr_to_expr.ml
+++ b/src/smtml/dolmenexpr_to_expr.ml
@@ -72,6 +72,276 @@ module Builtin = struct
       (DTy.arrow [ string_ty ] float64_ty)
 end
 
+module DolmenIntf = struct
+  include DTerm
+
+  type ty = DTy.t
+
+  type term = DTerm.t
+
+  type func_decl = DTerm.Const.t
+
+  let caches_consts = true
+
+  let true_ = DTerm._true
+
+  let false_ = DTerm._false
+
+  let int i = DTerm.Int.mk (string_of_int i)
+
+  let real r = DTerm.Real.mk (string_of_float r)
+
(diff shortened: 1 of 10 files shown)


src/smtml/mappings.ml
```diff
@@ -66,7 +66,7 @@ module Make (M_with_make : M_with_make) : S_with_fresh = struct
 
     let make_symbol (ctx : symbol_ctx) (s : Symbol.t) : symbol_ctx * M.term =
       let name = match s.name with Simple name -> name | _ -> assert false in
-      if M.caches_consts then
+      if not M.caches_consts then
```
filipeom, line 69:
```suggestion
      if not M.caches_consts then
```

I think this should remain without the `not` no? If the solver doesn't cache consts we want to do the lookup in the else branch to ensure we're not creating fresh symbols when the have the same name

hra687261: Right I assumed that it was the mapping that was meant to cache the consts, I will fix it

src/smtml/dolmenexpr_to_expr.ml
```diff
@@ -72,6 +72,276 @@ module Builtin = struct
...
+
+    let apply f tl = DTerm.apply_cst f [] tl
+  end
+
+  module Smtlib = struct
+    let pp ?name:_ ?logic:_ ?status:_ = Fmt.list DTerm.print
+  end
+end
+
 let tty_of_etype (e : Ty.t) : DTy.t =
```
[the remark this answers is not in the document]
hra687261: We don't use it yet for Alt-Ergo, so Alt-Ergo still uses that code. 
I will also do it for Alt-Ergo in a separate PR, I was going to wait for #326 to be merged, but I think I will just rebase on it.

filipeom:
Thank you so much!

PR #336, opened 2025-05-06
Normalize expressions for better term hash-consing and cache/constraint independance

felixL-K:
This PR adds basic normalization rules in Expr to improve term equivalence and hash-consing:

- not (a = b) → (a ≠ b)
- a > b → b < a
- a ≥ b → b ≤ a

These are handled directly in Expr.unop and Expr.relop.
Let me know if this is the right place for them or if you'd prefer these moved to a separate normalization step.

### DIFF
--- a/src/smtml/expr.ml
+++ b/src/smtml/expr.ml
@@ -328,6 +328,7 @@ let unop ty op hte =
   | _, Val v -> value (Eval.unop ty op v)
   | Not, Unop (_, Not, hte') -> hte'
   | Not, Relop (t, Eq, a,b) -> make (Relop (t, Ne, a, b))
+  | Not, Relop (t, Ne, a,b) -> make (Relop (t, Eq, a, b))
   | Neg, Unop (_, Neg, hte') -> hte'
   | Trim, Cvtop (Ty_real, ToString, _) -> hte
   | Head, List (hd :: _) -> hd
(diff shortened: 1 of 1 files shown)


src/smtml/expr.ml
```diff
@@ -424,7 +425,9 @@ let rec relop ty op hte1 hte2 =
     if Int32.equal b1 b2 then relop Ty_bool Eq os1 os2 else value False
   | Ne, Ptr { base = b1; offset = os1 }, Ptr { base = b2; offset = os2 } ->
     if Int32.equal b1 b2 then relop Ty_bool Ne os1 os2 else value True
-  | ( (LtU | LeU | GtU | GeU)
+  | GtU, _, _ -> relop ty LtU hte2 hte1
+  | GeU, _, _ -> relop ty LeU hte2 hte1
```
filipeom, line 429:
```suggestion
  | Gt, _, _ -> relop ty Lt hte2 hte1
  | GtU, _, _ -> relop ty LtU hte2 hte1
  | Ge, _, _ -> relop ty Le hte2 hte1
  | GeU, _, _ -> relop ty LeU hte2 hte1
``` 
It's missing for signed operations as well. 

But here you should leave the pattern match for the pointer compassion as is, because it will be resolved concretely. And, just to be safe, add these rules just before the default case: `| _, _, _ -> raw_relop ty op hte1 hte2` on line 446.

zapashcanon:
Note that, more than hash-consing, this is especially useful to increase the cache/constraint independance mechanism :)

filipeom (approved):
Thank you so much for this! Really appreciate the contribution!

PR #338, opened 2025-05-07
Normalize and simplify

felixL-K:
Trying to normalize two things but it make errors is the test suite : 

`smtml: internal error, uncaught exception:
        Failure("Bool: Unsupported Z3 unop operator \"is_nan\"")`

`smtml: internal error, uncaught exception:
        Failure("Bool: Unsupported Z3 binop operator \"sub\"")`

### DIFF
--- a/src/smtml/expr.ml
+++ b/src/smtml/expr.ml
@@ -325,8 +325,7 @@ let unop ty op hte =
   | Ty.Unop.(Regexp_loop _ | Regexp_star), _ -> raw_unop ty op hte
   | _, Val v -> value (Eval.unop ty op v)
   | Not, Unop (_, Not, hte') -> hte'
-  | Not, Relop (t, Eq, a, b) -> make (Relop (t, Ne, a, b))
-  | Not, Relop (t, Ne, a, b) -> make (Relop (t, Eq, a, b))
+  | Not, Relop (t, Eq, a,b) -> Hc.hashcons (Relop (t, Ne, a, b))
   | Neg, Unop (_, Neg, hte') -> hte'
   | Trim, Cvtop (Ty_real, ToString, _) -> hte
   | Head, List (hd :: _) -> hd
@@ -427,6 +426,8 @@ let rec relop ty op hte1 hte2 =
     if Int32.equal b1 b2 then relop Ty_bool Eq os1 os2 else value False
   | Ne, Ptr { base = b1; offset = os1 }, Ptr { base = b2; offset = os2 } ->
     if Int32.equal b1 b2 then relop Ty_bool Ne os1 os2 else value True
+  | GtU, _, _ -> relop ty LtU hte2 hte1
+  | GeU, _, _ -> relop ty LeU hte2 hte1
   | ( (LtU | LeU)
     , Ptr { base = b1; offset = os1 }
     , Ptr { base = b2; offset = os2 } ) ->
--- a/src/smtml/expr.ml
+++ b/src/smtml/expr.ml
@@ -316,8 +316,8 @@ let normalize_eq_or_ne op (ty', e1, e2) =
   match ty with
   | Ty_bitv a ->
      let binop = make (Binop (ty, Sub, e1, e2)) in
-     let zero = if a=32 then make (Val (Num (I32 0l)))
-                else make (Val (Num (I64 0L)))
+     let zero = if a=32 then make (Val (Bitv (Bitvector.of_int32 0l)))
+                else make (Val (Bitv (Bitvector.of_int64 0L)))
      in
       make_relop binop zero
   | Ty_fp 32 ->
@@ -405,8 +405,8 @@ let rec binop ty op hte1 hte2 =
   | Sub, Binop (ty, Sub, x, { node = Val v1; _ }), Val v2 ->
     let v = value (Eval.binop ty Add v1 v2) in
     raw_binop ty Sub x v
-  | Mul, Val (Num (I32 1l)), _ -> hte2
-  | Mul, _, Val (Num (I32 1l)) -> hte1
+  | Mul, Val (Value.Bitv bv), _ when Bitvector.eq_one bv -> hte2
+  | Mul, _, Val (Value.Bitv bv) when Bitvector.eq_one bv -> hte1
   | Mul, Binop (ty, Mul, x, { node = Val v1; _ }), Val v2 ->
     let v = value (Eval.binop ty Mul v1 v2) in
     raw_binop ty Mul x v
--- a/src/smtml/expr.ml
+++ b/src/smtml/expr.ml
@@ -326,6 +326,7 @@ let unop ty op hte =
   | _, Val v -> value (Eval.unop ty op v)
   | Not, Unop (_, Not, hte') -> hte'
   | Not, Relop (t, Eq, a,b) -> make (Relop (t, Ne, a, b))
+  | Not, Relop (t, Ne, a,b) -> make (Relop (t, Eq, a, b))
   | Neg, Unop (_, Neg, hte') -> hte'
   | Trim, Cvtop (Ty_real, ToString, _) -> hte
   | Head, List (hd :: _) -> hd
@@ -426,8 +427,6 @@ let rec relop ty op hte1 hte2 =
     if Int32.equal b1 b2 then relop Ty_bool Eq os1 os2 else value False
   | Ne, Ptr { base = b1; offset = os1 }, Ptr { base = b2; offset = os2 } ->
     if Int32.equal b1 b2 then relop Ty_bool Ne os1 os2 else value True
-  | GtU, _, _ -> relop ty LtU hte2 hte1
-  | GeU, _, _ -> relop ty LeU hte2 hte1
   | ( (LtU | LeU)
     , Ptr { base = b1; offset = os1 }
     , Ptr { base = b2; offset = os2 } ) ->
--- a/src/smtml/expr.ml
+++ b/src/smtml/expr.ml
@@ -405,8 +405,8 @@ let rec binop ty op hte1 hte2 =
   | Sub, Binop (ty, Sub, x, { node = Val v1; _ }), Val v2 ->
     let v = value (Eval.binop ty Add v1 v2) in
     raw_binop ty Sub x v
-  | Mul, Val (Value.Bitv bv), _ when Bitvector.eq_one bv -> hte2
-  | Mul, _, Val (Value.Bitv bv) when Bitvector.eq_one bv -> hte1
+  | Mul, Val (Bitv bv), _ when Bitvector.eq_one bv -> hte2
+  | Mul, _, Val (Bitv bv) when Bitvector.eq_one bv -> hte1
   | Mul, Binop (ty, Mul, x, { node = Val v1; _ }), Val v2 ->
     let v = value (Eval.binop ty Mul v1 v2) in
     raw_binop ty Mul x v
--- a/src/smtml/expr.ml
+++ b/src/smtml/expr.ml
@@ -303,42 +303,24 @@ let exists vars body = binder Exists vars body
 
 let raw_unop ty op hte = make (Unop (ty, op, hte)) [@@inline]
 
-let is_val_real e =
-  match view e with
-  | Val (Real x) -> Float.is_finite x
-  | _ -> false
-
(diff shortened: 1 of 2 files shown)


src/smtml/expr.ml
```diff
@@ -187,12 +187,13 @@ let get_symbols (hte : t list) =
 let negate_relop (hte : t) : (t, string) Result.t =
   let e =
     match view hte with
-    | Relop (ty, Eq, e1, e2) -> Ok (Relop (ty, Ne, e1, e2))
+    | Relop (ty, Eq, e1, e2) -> (* Ok (Relop (ty, Ne, e1, e2)) *)
+       Ok (Relop (ty, Ne, make (Binop (ty,Sub,e1,e2)), make (Val (Num (I32 0l)))))
```
hra687261, line 191: Are you sure that `I32 0l` is correct? What if `e1` and `e2` are of the type `I64` or anything else?

Assuming that `ty` in the pattern matching is the polymorphic equality's type parameter, `(ty,Sub,e1,e2)` is correct. But if that assumption is wrong, and `ty` is actually the type of the resulting expression, then the `ty` in the pattern matching is a boolean and the `ty` in  `(ty,Sub,e1,e2)` should be the type of `e1` (or `e2` which should be the same).

src/smtml/expr.ml
```diff
@@ -420,11 +425,16 @@ let rec relop ty op hte1 hte2 =
   | Ne, _, Val (App (`Op "symbol", [ Str _ ]))
   | Ne, Val (App (`Op "symbol", [ Str _ ])), _ ->
     value True
+  | Eq, (Symbol s1 as e1), (Symbol s2 as e2) ->
+    if Symbol.equal s1 s2 then
+      make (Unop (Ty_bool, Not, make (Unop (Ty_bool, Is_nan, make e1))))
```
hra687261, line 430: Related to the previous comment in `Unop (ty, _, _)`, is `ty` the resulting type or the type parameter, because if it is the type parameter then  `Unop (Ty_bool, Is_nan, make e1)` is incorrect, type of `e1` should be used instead of `Ty_bool`.
Since the type is not an option, and there are operations which are not polymorphic (like `Not`) I assume that it is the type of the resulting term from the application.

src/smtml/expr.ml
```diff
@@ -420,11 +425,16 @@ let rec relop ty op hte1 hte2 =
...
+    else
+      make (Relop (Ty_fp 32, Eq, make e1, make e2))
```
hra687261, line 432: Same here again. Here it looks like you assume that the type is a type parameter and not the type of the resulting expression, which is contrary to the previous case.

hra687261 (commented):
The issue probably comes from the usage of the types in the operations. It seems that the type in the operation tuples is the type of the resulting expression, but it seems that in some cases, it is assumed that it is the type parameter of the operation, which seems to be incorrect.

src/smtml/expr.ml
filipeom: Yeah, maybe you can use the function `ty` to get the type of the expressions?. (also, is this normalization safe for floats? Float.nan != 0. so it might be?)

src/smtml/expr.ml
```diff
@@ -366,6 +369,8 @@ let rec binop ty op hte1 hte2 =
   | Sub, Binop (ty, Sub, x, { node = Val v1; _ }), Val v2 ->
     let v = value (Eval.binop ty Add v1 v2) in
     raw_binop ty Sub x v
+  | Mul, Val (Num (I32 1l)), _ -> hte2
+  | Mul, _, Val (Num (I32 1l)) -> hte1
```
filipeom, line 373:
```suggestion
  | Mul, Val (Bitv bv), _ when Bitvector.eq_one bv -> hte2
  | Mul, _, Val (Bitv bv) when Bitvector.eq_one bv -> hte1
```

Rebase for these changes

src/smtml/expr.ml
```diff
@@ -440,6 +450,10 @@ let rec relop ty op hte1 hte2 =
     let base = Eval.binop (Ty_bitv 32) Add (Num (I32 base)) o in
     value (if Eval.relop ty op base n then True else False)
   | op, List l1, List l2 -> relop_list op l1 l2
+  | Gt, _, _ -> relop ty Lt hte2 hte1
+  | GtU, _, _ -> relop ty LtU hte2 hte1
+  | Ge, _, _ -> relop ty Le hte2 hte1
+  | GeU, _, _ -> relop ty LeU hte2 hte1
```
filipeom, line 456: This was already merged. But maybe, to avoid accidentally constructing `Gt|Ge` expressions we could remove this datatype from `Ty.Relop.t`? (It would perhaps best be addressed in a separate PR)

src/smtml/expr.ml
hra687261: For floats, I am not sure, there are probably tricky corner cases, one i can think of: 
- `+inf = +inf` is true
- `+inf - +inf = 0` is false because `+inf - +inf` is NaN

felixL-K:
Thank you very much for all your feedback.
I’ve made the necessary corrections, and I now understand that handling floating-point values is going to be tricky to normalize properly.
For now, I’ve added checks to ensure the values are not NaN before applying certain transformations, but I suspect this won’t be sufficient in the long run.

src/smtml/expr.ml
```diff
@@ -423,6 +461,11 @@ let rec relop ty op hte1 hte2 =
   | Ne, _, Val (App (`Op "symbol", [ Str _ ]))
   | Ne, Val (App (`Op "symbol", [ Str _ ])), _ ->
     value True
+  | Eq, (
+    (Symbol ({ty=Ty_fp prec1; _} as s1)) as e1),
+    (Symbol ({ty=Ty_fp prec2; _} as s2))
+       when prec1 = prec2 && Symbol.equal s1 s2 ->
+     make (Unop (Ty_bool, Not, make (Unop (Ty_fp prec1, Is_nan, make e1))))
```
filipeom, line 468:
```suggestion
     raw_unop Ty_bool Not (raw_unop (Ty_fp prec1) Is_nan hte1)
```

Uses the `raw_unop` operator and `hte1` instead of `e1` to doing a hashcons lookup

src/smtml/expr.ml
```diff
@@ -320,13 +303,66 @@ let exists vars body = binder Exists vars body
...
+let is_val_real e =
+  match view e with
+  | Val (Real x) -> Float.is_finite x
+  | _ -> false
+
+let normalize_eq_or_ne op (ty', e1, e2) =
+  let is_real = is_val_real e1 && is_val_real e2 in
+  let make_relop lhs rhs = Relop (ty', op, lhs, rhs) in
+  let ty,ty2 = ty e1,ty e2 in
+  if Ty.equal ty ty2 then
```
filipeom, line 315:
```suggestion
  assert (Ty.equal ty ty2);
```
At some point in the future, I plan to ensure type safety during expression construction. So, I expect this invariant to always hold, so we can replace the `if` with an `assert` as a sanity check.

src/smtml/expr.ml
```diff
@@ -320,13 +303,66 @@ let exists vars body = binder Exists vars body
...
+  match ty with
+  | Ty_bitv a ->
+     let binop = make (Binop (ty, Sub, e1, e2)) in
+     let zero = if a=32 then make (Val (Bitv (Bitvector.of_int32 0l)))
+                else make (Val (Bitv (Bitvector.of_int64 0L)))
+     in
+      make_relop binop zero
```
filipeom, line 322:
```suggestion
  | Ty_bitv m ->
     let binop = make (Binop (ty, Sub, e1, e2)) in
     let zero = make (Val (Bitv (Bitvector.make Z.zero m))) in
      make_relop binop zero
```

No need for an if-statement, just say you want a zero with the size of the `Ty_bitv`

src/smtml/expr.ml
```diff
@@ -320,13 +303,66 @@ let exists vars body = binder Exists vars body
...
+          make (Binop (ty, Sub, e1, e2)), make (Val (Num (F32 0l)))
+        else e1, e2
+      in
+      make_relop lhs rhs
+  | Ty_fp 64 ->
+      let lhs, rhs = if is_real then
+          make (Binop (ty, Sub, e1, e2)), make (Val (Num (F64 0L)))
+        else e1, e2
+      in
+      make_relop lhs rhs
```
filipeom, line 334: From what I understand, you're using `is_real` to check whether `Val (Real x)` satisfies `Float.is_finite x`. In the case where the type is `Ty_fp _`, you can never have a numeric `Val (Real _)`, so the `if` statement is redundant, it will always take the false branch. A more appropriate pattern match would be the following:


```suggestion
  | Ty_real -> make (Binop (ty, Sub, e1, e2)), make (Val (Real 0.)))
```

This only applies to arithmetic reals, and it's safe to remove the `is_real` check because real arithmetic does not include special floating-point values like `inf`.

src/smtml/expr.ml
```diff
@@ -320,13 +303,66 @@ let exists vars body = binder Exists vars body
...
+  | Ty_int ->
+     let binop = make (Binop (ty, Sub, e1, e2)) in
+      let zero = make (Val (Int (Int.zero))) in
+      make_relop binop zero
+  | _ -> Fmt.failwith "Invalid relop (%a) : (%a) (%a) (%a)"
+           Ty.pp ty' pp e1 Ty.Relop.pp op pp e2
```
filipeom, line 340:
```suggestion
  | _ -> make_relop e1 e2
```

The base case should be `make_relop` because you might get expressions such as:

```ocaml
Relop (Ty_bool, Eq, Val (Str "abc"), Val (Str "cda"))
```

Which you don't want to normalize

filipeom (commented):
Just a few comments on `normalize_eq_or_ne` which should be taken care of, and few style comments.

Also, if you don't have already, can you install `ocamlformat` (with `opam install ocamlformat`) and run `dune fmt`?

filipeom:
Thanks!

filipeom (approved):

src/smtml/expr.ml
```diff
@@ -320,13 +303,48 @@ let exists vars body = binder Exists vars body
...
+    let binop = make (Binop (ty, Sub, e1, e2)) in
+    let zero = make (Val (Real 0.)) in
+    make_relop binop zero
+  | _ -> make_relop e1 e2
+
+let negate_relop (hte : t) : t =
+  let e =
+    match view hte with
+    | Relop (ty, Eq, e1, e2) -> normalize_eq_or_ne Ne (ty, e1, e2)
+    | Relop (ty, Ne, e1, e2) -> normalize_eq_or_ne Eq (ty, e1, e2)
```
zapashcanon, line 329: I think you could have written this:

```ocaml
    | Relop (ty, ((Eq|Ne) as op), e1, e2) -> normalize_eq_or_ne op (ty, e1, e2)
```

:D

src/smtml/expr.ml
```diff
@@ -423,6 +443,11 @@ let rec relop ty op hte1 hte2 =
   | Ne, _, Val (App (`Op "symbol", [ Str _ ]))
   | Ne, Val (App (`Op "symbol", [ Str _ ])), _ ->
     value True
+  | ( Eq
+    , Symbol ({ ty = Ty_fp prec1; _ } as s1)
+    , Symbol ({ ty = Ty_fp prec2; _ } as s2) )
+    when prec1 = prec2 && Symbol.equal s1 s2 ->
```
zapashcanon, line 449: Isn't `Symbol.equal` already comparing `prec1` and `prec2` ?

src/smtml/expr.ml
felixL-K: Good point, I changed it in my new normalize branch.

src/smtml/expr.ml
felixL-K: It doesn't look like it : Symbol.compare uses Ty.compare on the Ty_fp, but that doesn't compare the size, if I understand correctly.

zapashcanon: This is a little bit surprising, cc @filipeom who may have some insight about it :D

filipeom: I'm also certain that `Ty.compare` will compare the size of the floating points. So, this should hold `Ty.compare (Ty_fp 32) (Ty_fp 32) = 0`. I think it's safe to do what @zapashcanon is saying.

PR #342, opened 2025-05-12
Handle unknown smt2 functions as apps

### LINKED ISSUE #340: Provide a way to handle unknown term apps
Since not all of the smtlib term apps are supported it's relatively easy to get the `"could not parse term app: ..."` error. It makes it impossible to work with `.smt2` files that have some unsupported apps in them.

Though in general support parsing such terms might be handy. E.g. one might wanna transform the unknown term apps into something that is supported. It'd nice to provide such way not to fail when parsing `smt2`.

georgiy-belyanin:
This patch makes smt2 implementation not fail on unknown function symbols but instead parse them as apps (i.e. `Expr.App`).

It allows one to parse smt2 and process such apps on its own extending the implemented symbols.

Closes #340

### DIFF
--- a/src/smtml/smtlib.ml
+++ b/src/smtml/smtlib.ml
@@ -113,7 +113,7 @@ module Term = struct
 
   let apply ?loc (id : t) (args : t list) : t =
     match Expr.view id with
-    | Symbol { namespace = Term; name = Simple name; _ } -> (
+    | Symbol ({ namespace = Term; name = Simple name; _ } as symbol) -> (
       match (name, args) with
       | "-", [ a ] -> Expr.raw_unop Ty_none Neg a
       | "not", [ a ] -> Expr.raw_unop Ty_bool Not a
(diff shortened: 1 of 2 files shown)


georgiy-belyanin:
For some reason CI has failed during installing dependencies. Could you, please, re-trigger it? Thank you.

src/smtml/smtlib.ml
```diff
@@ -286,8 +286,7 @@ module Term = struct
         Expr.raw_unop Ty_regexp (Regexp_loop (i1, i2)) a
       | _ ->
         Fmt.failwith "%acould not parse indexed app: %a" pp_loc loc Expr.pp id )
-    | Symbol id ->
-      Fmt.failwith "%acould not parse app: %a" pp_loc loc Symbol.pp id
+    | Symbol id -> Expr.app id []
```
filipeom, line 289:
```suggestion
    | Symbol id -> Expr.app id args
```

Maybe here we should also pass the arguments to the app?

georgiy-belyanin: Yes, it was just a mistake. Fixed. Thank you for pointing out & force-pushed the fixes.

filipeom (approved):
This looks good to go for me! 

Do you also need stuff like uninterpreted functions to be supported in the solver? This PR now allows to do that I think. Stuff like this:

```smt2
(declare-fun f ((_ FloatingPoint 8 24)) (_ BitVec 32))
(declare-const x (_ FloatingPoint 8 24))
(assert (= (f x) #x0000002a))
(check-sat)
(get-model)
```

georgiy-belyanin:
That's cool. Thank you for taking a look and for a nice example. Both uninterpreted functions and custom-defined function symbols are really handy. We're currently researching automata-based SMT algorithms for exponents. Frontend that also allows one to communicate with other solvers is really great solution for experimenting.

filipeom:
> That's cool. Thank you for taking a look and for a nice example. Both uninterpreted functions and custom-defined function symbols are really handy. We're currently researching automata-based SMT algorithms for exponents. Frontend that also allows one to communicate with other solvers is really great solution for experimenting.

Great! I'll make a separate PR to connect stuff like this to the solver :)

PR #347, opened 2025-05-14
Use mappings.ml for Alt-Ergo's mappings

hra687261:


filipeom (approved):
Thanks! I'm merging this, but can you briefly explain why the AltErgo mappings need the symbol map to be passed to the `add`, `check`, and `eval` functions? It probably has something to do with model generation right?

hra687261:
Yes, one of the reasons is model generation, more precisely the `eval` function that computes a new term by substituting variables with their assigned values. I currently use Dolmen's model evaluation to do that (because we can't do it on Alt-Ergo terms), so since we don't have a mapping from alt-ergo's symbols to dolmen symbols, it is simpler to just get the symbols.
Also, Alt-Ergo's API is not really one, so we just use the same functions used after parsing/typechecking terms and these functions expect every user provided symbol to be declared first before being used. I think that can be changed but it needs to be changed only when Alt-Ergo is used as a library from the API, because we don't want that behavior in the general case.

filipeom:
> Yes, one of the reasons is model generation, more precisely the `eval` function that computes a new term by substituting variables with their assigned values. I currently use Dolmen's model evaluation to do that (because we can't do it on Alt-Ergo terms), so since we don't have a mapping from alt-ergo's symbols to dolmen symbols, it is simpler to just get the symbols. Also, Alt-Ergo's API is not really one, so we just use the same functions used after parsing/typechecking terms and these functions expect every user provided symbol to be declared first before being used. I think that can be changed but it needs to be changed only when Alt-Ergo is used as a library from the API, because we don't want that behavior in the general case.

Okok, thanks for the explanation!

PR #348, opened 2025-05-16
 Normalize and simplify

felixL-K:
Trying to normalize some expr

### DIFF
--- a/src/smtml/expr.ml
+++ b/src/smtml/expr.ml
@@ -371,9 +371,6 @@ let rec binop ty op hte1 hte2 =
   | Add, Binop (ty, Add, x, { node = Val v1; _ }), Val v2 ->
     let v = value (Eval.binop ty Add v1 v2) in
     raw_binop ty Add x v
-  | And, Binop (ty, And, x, { node = Val v1; _ }), Val v2 ->
-    let v = value (Eval.binop ty And v1 v2) in
-    raw_binop ty And x v
   | Sub, Binop (ty, Sub, x, { node = Val v1; _ }), Val v2 ->
     let v = value (Eval.binop ty Add v1 v2) in
     raw_binop ty Sub x v


felixL-K:
I'm trying to make some smt2 examples to test the conversion simplifications: 

`(set-logic QF_BVFP)
(define-fun x () (_ BitVec 32))
(define-fun y () (_ FloatingPoint 8 24))
(assert (= x ((_ fp.to_ubv 32) roundNearestTiesToEven y)))
(check-sat)`
and
`(set-logic QF_BVFP)
(define-fun x () (_ BitVec 32))
(define-fun y () (_ FloatingPoint 8 24))
(assert (= y (fp #b0 #b01011100111 #b1111110000101111001101101000100010000000100100100111)))
(assert (= x ((_ fp.to_ubv 32) roundNearestTiesToEven y)))
(check-sat)`

But I don't manage to get them to work, I would be very grateful if someone can help me correct one of the examples.

src/smtml/expr.ml
```diff
@@ -501,6 +516,26 @@ let cvtop ty op hte =
   | Ty.Cvtop.String_to_re, _ -> raw_cvtop ty op hte
   | _, Val v -> value (Eval.cvtop ty op v)
   | String_to_float, Cvtop (Ty_real, ToString, real) -> real
+  | Reinterpret_float, Cvtop (Ty_real, Reinterpret_int, e1) -> e1
+  | Reinterpret_float, Cvtop (Ty_fp 32, Reinterpret_int, e1) -> e1
+  | Reinterpret_float, Cvtop (Ty_fp 64, Reinterpret_int, e1) -> e1
```
zapashcanon, line 521: I think the size of e1 should be checked in the two previous examples? And also in the following ones

src/smtml/expr.ml
```diff
@@ -501,6 +516,26 @@ let cvtop ty op hte =
...
+  | Reinterpret_int, Cvtop (Ty_int, Reinterpret_float, e1) -> e1
+  | Reinterpret_int, Cvtop (Ty_bitv 32, Reinterpret_float, e1) -> e1
+  | Reinterpret_int, Cvtop (Ty_bitv 64, Reinterpret_float, e1) -> e1
+  | ToString, Cvtop (_, OfString, e1) -> e1
+  | OfString, Cvtop (_, ToString, e1) -> e1
+  | ToBool, Cvtop (_, OfBool, e1) -> e1
+  | OfBool, Cvtop (_, ToBool, e1) -> e1
+  | PromoteF32, Cvtop (_, DemoteF64, e1) -> e1
```
zapashcanon, line 529: I am not sure if this is correct. When demoting the F64, you get a F32 where you lost some precision, and thus, the resulting F64 is not the same as the initial one?

src/smtml/expr.ml
```diff
@@ -501,6 +516,26 @@ let cvtop ty op hte =
...
+  | DemoteF64, Cvtop (_, PromoteF32, e1) -> e1
+  | Reinterpret_int, Cvtop (_, Reinterpret_float, e1) -> e1
+  | Reinterpret_float, Cvtop (_, Reinterpret_int, e1) -> e1
+  | Zero_extend 0, _ -> hte
+  | Sign_extend 0, _ -> hte
+  | String_from_code, Cvtop (_, String_to_code, e1) -> e1
+  | String_to_code, Cvtop (_, String_from_code, e1) -> e1
+  | String_to_int, Cvtop (_, String_from_int, e1) -> e1
+  | String_from_int, Cvtop (_, String_to_int, e1) -> e1
```
zapashcanon, line 538: I am not sure about the semantics of String_from_it but, if it can fail, then I'd say these are not correct

src/smtml/expr.ml
```diff
@@ -358,6 +357,12 @@ let raw_binop ty op hte1 hte2 = make (Binop (ty, op, hte1, hte2)) [@@inline]
 let rec binop ty op hte1 hte2 =
   match (op, view hte1, view hte2) with
   | Ty.Binop.(String_in_re | Regexp_range), _, _ -> raw_binop ty op hte1 hte2
+  | Ty.Binop.String_contains, _, _ when equal hte2 (make (Val (Str ""))) ->
+    make (Val True)
+  | Ty.Binop.String_prefix, _, _ when equal hte2 (make (Val (Str ""))) ->
+    make (Val True)
+  | Ty.Binop.String_suffix, _, _ when equal hte2 (make (Val (Str ""))) ->
+    make (Val True)
```
filipeom, line 365:
```suggestion
  | Ty.Binop.String_contains, _, Val (Str "") -> make (Val True)
  | Ty.Binop.String_prefix, _, Val (Str "") -> make (Val True)
  | Ty.Binop.String_suffix, _, Val (Str "") -> make (Val True)
```

I think we can do this, it removes the necessity of using the `when`

src/smtml/expr.ml
filipeom: You think so? I was actually just going to suggest something like:

```suggestion
  | Reinterpret_float, Cvtop (Ty_fp n, Reinterpret_int, e) -> 
    assert (match ty with Ty_bitv m -> n = m | _ -> false);
    e
```
I think it would be enough and we could reduce the two pattern matches on `Ty_fp` into a single one

src/smtml/expr.ml
```diff
@@ -501,6 +516,26 @@ let cvtop ty op hte =
   | Ty.Cvtop.String_to_re, _ -> raw_cvtop ty op hte
   | _, Val v -> value (Eval.cvtop ty op v)
   | String_to_float, Cvtop (Ty_real, ToString, real) -> real
+  | Reinterpret_float, Cvtop (Ty_real, Reinterpret_int, e1) -> e1
+  | Reinterpret_float, Cvtop (Ty_fp 32, Reinterpret_int, e1) -> e1
+  | Reinterpret_float, Cvtop (Ty_fp 64, Reinterpret_int, e1) -> e1
+  | Reinterpret_int, Cvtop (Ty_int, Reinterpret_float, e1) -> e1
+  | Reinterpret_int, Cvtop (Ty_bitv 32, Reinterpret_float, e1) -> e1
+  | Reinterpret_int, Cvtop (Ty_bitv 64, Reinterpret_float, e1) -> e1
```
filipeom, line 524:
```suggestion
  | Reinterpret_int, Cvtop (Ty_bitv m, Reinterpret_float, e) -> 
    assert (match ty with Ty_fp n -> n = m | _ -> false);
    e
```

src/smtml/expr.ml
filipeom: Yes, these two I'm not sure the round-trip is safe

src/smtml/expr.ml
```diff
@@ -501,6 +516,26 @@ let cvtop ty op hte =
...
+  | ToString, Cvtop (_, OfString, e1) -> e1
+  | OfString, Cvtop (_, ToString, e1) -> e1
+  | ToBool, Cvtop (_, OfBool, e1) -> e1
+  | OfBool, Cvtop (_, ToBool, e1) -> e1
+  | PromoteF32, Cvtop (_, DemoteF64, e1) -> e1
+  | DemoteF64, Cvtop (_, PromoteF32, e1) -> e1
+  | Reinterpret_int, Cvtop (_, Reinterpret_float, e1) -> e1
+  | Reinterpret_float, Cvtop (_, Reinterpret_int, e1) -> e1
```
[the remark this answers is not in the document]
filipeom: I think these are probably never allowed. I'm working on a PR that should never allow this. It will probably remove the need for this pattern match.

src/smtml/expr.ml
filipeom: I'd also have to refresh my memory on these

filipeom (commented):
Some comments below. Regarding the issue with testing using these examples:

```smt2
(set-logic QF_BVFP)
(declare-fun x () (_ BitVec 32))
(declare-fun y () (_ FloatingPoint 8 24))
(assert (= x ((_ fp.to_ubv 32) roundNearestTiesToEven y)))
(check-sat)
```

These are not working because the parser fails? I'm going to be working on fixing some issues in the parser. So I can take a look at these examples. 

In any case, I don't think using the smtlib to test simplifications is the most easy thing to do. I'd actually recommend writing unit tests here: https://github.com/formalsec/smtml/blob/main/test/unit/test_expr.ml

For example you could use these formulas here: for inspiration to try out some of the rules you wrote. For example, for these:

```
(f32.add
 (f32.add
   (f32.reinterpret_int (i32.reinterpret_float symbol_0)) 
   0.) 
 (f32.reinterpret_int (i32.reinterpret_float symbol_1))))
```

You could write:

```ocaml

let test_simplify_reinterpret_int _ =
  let open Infix in
  let sym_0 = symbol "symbol_0" (Ty_fp 32) in
  let lhs =
    Expr.binop (Ty_fp 32) Add 
      (Expr.cvtop (Ty_fp 32) Reinterpret_float 
        (Expr.cvtop (Ty_bitv 32) Reinterpret_int sym_0))
      (float32 0.)
   in
   let sym_1 = symbol "symbol_1" (Ty_fp 32) in
   let rhs =
      Expr.cvtop (Ty_fp 32) Reinterpret_float 
        (Expr.cvtop (Ty_bitv 32) Reinterpret_int sym_0)
   in
   let result = Expr.binop (Ty_fp 32) Add lhs rhs in 
   let expected = Expr.binop (Ty_fp 32) Add sym_0 sym_1 in
   check result expected
```

src/smtml/expr.ml
zapashcanon: Yes, looks good too

src/smtml/expr.ml
```diff
@@ -357,12 +357,9 @@ let raw_binop ty op hte1 hte2 = make (Binop (ty, op, hte1, hte2)) [@@inline]
...
   | Ty.Binop.(String_in_re | Regexp_range), _, _ -> raw_binop ty op hte1 hte2
-  | Ty.Binop.String_contains, _, _ when equal hte2 (make (Val (Str ""))) ->
-    make (Val True)
-  | Ty.Binop.String_prefix, _, _ when equal hte2 (make (Val (Str ""))) ->
-    make (Val True)
-  | Ty.Binop.String_suffix, _, _ when equal hte2 (make (Val (Str ""))) ->
-    make (Val True)
+  | Ty.Binop.String_contains, _, Val (Str "") -> make (Val True)
+  | Ty.Binop.String_prefix, _, Val (Str "") -> make (Val True)
+  | Ty.Binop.String_suffix, _, Val (Str "") -> make (Val True)
```
zapashcanon, line 362: You can merge the three cases:

```ocaml
 | (Ty.Binop.String_contains | Ty.Binop.String_prefix | Ty.Binop.String_suffix), _, Val (Str "") -> make (Val True)
 ```

File: src/smtml/expr.ml [hunk 8]
```diff
@@ -381,9 +383,18 @@ let rec binop ty op hte1 hte2 =
     hte1
   | (Add | Or), _, Val (Bitv bv) when Bitvector.eqz bv -> hte1
   | (And | Mul), _, Val (Bitv bv) when Bitvector.eqz bv -> hte2
+  | And, Val True, _ -> hte2
+  | And, _, Val True -> hte1
+  | And, Val False, _ | And, _, Val False -> make (Val False)
```
zapashcanon, line 388: `make (Val False)` is `hte1` right? (same for the next one)

src/smtml/expr.ml
```diff
@@ -381,9 +383,18 @@ let rec binop ty op hte1 hte2 =
...
+  | Or, Val True, _ | Or, _, Val True -> make (Val True)
+  | Or, Val False, _ -> hte2
+  | Or, _, Val False -> hte1
   | Add, Binop (ty, Add, x, { node = Val v1; _ }), Val v2 ->
     let v = value (Eval.binop ty Add v1 v2) in
     raw_binop ty Add x v
+  | And, Binop (ty, And, x, { node = Val v1; _ }), Val v2 ->
+    let v = value (Eval.binop ty And v1 v2) in
+    raw_binop ty And x v
```
zapashcanon, line 397: I don't understand how is this case different from the previous one ?

src/smtml/expr.ml
felixL-K: you mean I could do this : 
`
| And, Val False, _ -> hte1
| And, _, Val False -> hte2
`
Is it better than to recreate with `make` ?

zapashcanon: Yes I think it's better, I guess `make` is looking in the hashconsing table and everything and there's no need to do it here

src/smtml/expr.ml
```diff
@@ -382,6 +384,10 @@ let rec binop ty op hte1 hte2 =
     when Bitvector.eqz bv ->
     hte1
   | (Add | Or), _, Val (Bitv bv) when Bitvector.eqz bv -> hte1
+  | Add, _, Val (Real 0.) -> hte1
+  | Add, Val (Real 0.), _ -> hte2
+  | Add, _, Val (Num (F32 0l)) -> hte1
+  | Add, Val (Num (F32 0l)), _ -> hte2
```
zapashcanon, line 390: I don't think this is right. On floats `+0. + -0.` is going to give `+0.`. (I.e. the neutral element is `-0` and not `+0`.)

src/smtml/expr.ml
```diff
@@ -385,14 +385,15 @@ let rec binop ty op hte1 hte2 =
...
     hte1
   | (Add | Or), _, Val (Bitv bv) when Bitvector.eqz bv -> hte1
-  | Add, _, Val (Real 0.) -> hte1
-  | Add, Val (Real 0.), _ -> hte2
-  | Add, _, Val (Num (F32 0l)) -> hte1
-  | Add, Val (Num (F32 0l)), _ -> hte2
+  | Add, _, Val (Real -0.) -> hte1
+  | Add, Val (Real -0.), _ -> hte2
+  | Add, _, Val (Num (F32 -0l)) -> hte1
+  | Add, Val (Num (F32 -0l)), _ -> hte2
```
zapashcanon, line 391: Why not doing it for `F64` too?

src/smtml/expr.ml
```diff
@@ -383,10 +388,24 @@ let rec binop ty op hte1 hte2 =
...
   | (Add | Or), _, Val (Bitv bv) when Bitvector.eqz bv -> hte1
+  | Add, _, Val (Real -0.) -> hte1
+  | Add, Val (Real -0.), _ -> hte2
+  | Add, _, Val (Num (F32 -0l)) -> hte1
+  | Add, Val (Num (F32 -0l)), _ -> hte2
   | (And | Mul), _, Val (Bitv bv) when Bitvector.eqz bv -> hte2
+  | And, Val True, _ -> hte2
+  | And, _, Val True -> hte1
+  | And, Val False, _ -> hte1
+  | And, _, Val False -> hte2
```
hra687261, line 399:
```suggestion
  | And, Val False, _ | And, _, Val False -> make (Val False)
```
Anything `and false` results in `false`.

zapashcanon: Yes but here, `hte1` and `hte2` are the false expression! This is to avoid hash-consing again I guess?

src/smtml/expr.ml
```diff
@@ -385,14 +385,15 @@ let rec binop ty op hte1 hte2 =
...
+  | Add, Val (Real -0.), _ -> hte2
+  | Add, _, Val (Num (F32 -0l)) -> hte1
+  | Add, Val (Num (F32 -0l)), _ -> hte2
   | (And | Mul), _, Val (Bitv bv) when Bitvector.eqz bv -> hte2
   | And, Val True, _ -> hte2
   | And, _, Val True -> hte1
-  | And, Val False, _ | And, _, Val False -> make (Val False)
+  | And, Val False, _ -> hte1
+  | And, _, Val False -> hte2
   | Or, Val True, _ | Or, _, Val True -> make (Val True)
```
hra687261, line 400:
```suggestion
  | Or, Val True, _ -> hte1
  | Or, _, Val True -> hte2
```
This can be handled similarly to `And`, no?

src/smtml/expr.ml
```diff
@@ -383,10 +388,24 @@ let rec binop ty op hte1 hte2 =
...
+  | And, _, Val False -> hte2
+  | Or, Val True, _ | Or, _, Val True -> make (Val True)
+  | Or, Val False, _ -> hte2
+  | Or, _, Val False -> hte1
   | Add, Binop (ty, Add, x, { node = Val v1; _ }), Val v2 ->
     let v = value (Eval.binop ty Add v1 v2) in
     raw_binop ty Add x v
+  | And, Binop (ty, And, x, { node = Val v1; _ }), Val v2 ->
+    let v = value (Eval.binop ty And v1 v2) in
+    raw_binop ty And x v
```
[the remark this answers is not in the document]
hra687261: Indeed, but what I meant was that this case should be handled by the simplification of `And` with constants no? i.e. `(And (And cst v1) v2)` should be turned into `(And v1 v2)` or `(And cst v2)` by the `And` cases above. `(And cst v1)` should be simplified by the recursive call, I don't think we need to catch it here.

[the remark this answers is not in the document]
felixL-K: Ok I see now sorry, I remove this part.

PR #353, opened 2025-05-20
Make `rotate_{left|right}` operators indexed following the SMT-LIB spec

filipeom:
- Add a unary operator `Rotate_{left|right} of int` to model indexed rotation operators, as defined by the SMT-LIB semantics.

- Retain the binary variants of `rotate_{left|right}` using the operators `Ext_rotate_{left|right}`, where `Binop (Ext_rotate_left, t1, t2)` denotes the left rotation of `t1` by `t2` bits.


filipeom:
Reflecting on this, I don't know if I like this very much. We're repeating ourselves, two ways of creating the same thing

PR #360, opened 2025-05-30
Fix more parsing errors

filipeom:


filipeom:
ON the `QF_FP` benchmarks: 

- Passed: 40153/40406 (99.37%)

253 missing. But I think we can merge this

PR #369, opened 2025-06-07
Delay initial allocation costs of sorts with "lazy" sorts

filipeom:
@zapashcanon what do you think of this? Any opinions on the use of `lazy`?

### DIFF
--- a/src/smtml/bitwuzla_mappings.default.ml
+++ b/src/smtml/bitwuzla_mappings.default.ml
@@ -67,19 +67,19 @@ module Fresh_bitwuzla (B : Bitwuzla_cxx.S) : M = struct
   let exists _ = Fmt.failwith "Bitwuzla_mappings: exists not implemented"
 
   module Types = struct
-    let int = mk_bool_sort ()
+    let int = lazy (mk_bool_sort ())
 
-    let real = mk_bool_sort ()
+    let real = lazy (mk_bool_sort ())
 
-    let bool = mk_bool_sort ()
+    let bool = lazy (mk_bool_sort ())
 
-    let string = mk_bool_sort ()
+    let string = lazy (mk_bool_sort ())
 
     let bitv bitwidth = mk_bv_sort bitwidth
 
     let float ebits sbits = mk_fp_sort ebits sbits
 
-    let roundingMode = mk_rm_sort ()
+    let roundingMode = lazy (mk_rm_sort ())
 
     let ty t = Term.sort t
 
--- a/src/smtml/cvc5_mappings.default.ml
+++ b/src/smtml/cvc5_mappings.default.ml
@@ -65,19 +65,19 @@ module Fresh_cvc5 () = struct
   let exists _ = Fmt.failwith "Cvc5_mappings: exists not implemented"
 
   module Types = struct
-    let int = Sort.mk_int_sort tm
+    let int = lazy (Sort.mk_int_sort tm)
 
-    let real = Sort.mk_real_sort tm
+    let real = lazy (Sort.mk_real_sort tm)
 
-    let bool = Sort.mk_bool_sort tm
+    let bool = lazy (Sort.mk_bool_sort tm)
 
-    let string = Sort.mk_string_sort tm
+    let string = lazy (Sort.mk_string_sort tm)
 
     let bitv bitwidth = Sort.mk_bv_sort tm bitwidth
 
     let float ebits sbits = Sort.mk_fp_sort tm ebits sbits
 
-    let roundingMode = Sort.mk_rm_sort tm
+    let roundingMode = lazy (Sort.mk_rm_sort tm)
 
     let ty t = Term.sort t
 
--- a/src/smtml/dolmenexpr_to_expr.ml
+++ b/src/smtml/dolmenexpr_to_expr.ml
@@ -144,6 +144,16 @@ module DolmenIntf = struct
   module Types = struct
     include DTy
 
+    let int = lazy int
+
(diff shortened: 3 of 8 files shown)


zapashcanon:
I'm away so I only had a quick look but:

1. Currently `lazy` is not thread-safe in OCaml 5 so I am not sure this is correct!
2. Most of them look like very small allocations and I am not sure the use of `lazy` is actually worth it. For some that are not compiled to blocks, I am also almost sure this is going to make things worse (e.g. ```lazy `Typ```, which is actually not a block IIRC, you can check with ```Obj.is_block (Obj.repr `Typ)```). Maybe someone like @chambart or @lthls has an idea of when this can be beneficial.

filipeom:
Thanks for the response! 

> I'm away so I only had a quick look but:
> 
>     1. Currently `lazy` is not thread-safe in OCaml 5 so I am not sure this is correct!

Yeah, I was thinking about this as well. But I don’t think we need synchronization here, since each thread creates its own instance of the mappings using the corresponding solver mappings. So, each thread calls `Z3_mappings.Make ()` to use solvers in parallel, so `lazy` should memoise a separate value, right?

> 
>     2. Some of them look like very small allocations and I am not sure the use of `lazy` is actually worth it.

Yeah, maybe this is the biggest argument against this PR. I feel like I don't want to introduce these changes with the `lazy` values. But It also bothers me that we're doing unnecessary allocations for every solver. 

> For some that are not compiled to blocks, I am also almost sure this is going to make things worse (e.g. `` lazy `Typ ``, which is actually not a block IIRC, you can check with `` Obj.is_block (Obj.repr `Typ) ``). Maybe someone like @chambart or @lthls has an idea of when this can be beneficial.

These: `` lazy `Typ `` are actually from the nop mappings, so they shouldn't be used in practice when a solver is installed, I think.

lthls:
Note that `` lazy `Typ `` is not a block either. It is represented in the same way as `` `Typ ``. (This actually holds for most OCaml values: `lazy expr`, when forced, will eventually use the same representation as the final value of `expr`, unless `expr` is a lazy value itself or a float).

filipeom:
Ran with 5s timeout

```
tool1 had 592 reached and tool2 had 592 reached
tool1 had 618 timeout and tool2 had 619 timeout
tool1 had 001 nothing and tool2 had 001 nothing
tool1 had 004 other   and tool2 had 003 other
tool1 had 000 killed  and tool2 had 000 killed

tools have 583 reached tasks in common

tool1 had 009 tasks tool2 did not found
tool2 had 009 tasks tool1 did not found

on      commonly  reached tasks, tool 1 took 310.420000 sec. (mean 0.532453, median 0.210000, min 0.090000, max 4.580000) and tool 2 took 305.760000 sec. (mean 0.524460, median 0.210000, min 0.090000, max 4.760000)
on *not commonly* reached tasks, tool 1 took 39.350000 sec. (mean 4.372222, median 4.700000, min 2.060000, max 4.960000) and tool 2 took 32.420000 sec. (mean 3.602222, median 3.890000, min 1.840000, max 4.830000)

among tasks reached only by tool 1, tool 2 replied 000 nothing, 009 timeout, 00 other and 00 killed
among tasks reached only by tool 2, tool 1 replied 000 nothing, 008 timeout, 01 other and 00 killed
tasks solved only by tool 1:
  testcomp/sv-benchmarks/c/xcsp/aim-100-1-6-sat-3.c
  testcomp/sv-benchmarks/c/xcsp/AllInterval-013.c
  testcomp/sv-benchmarks/c/xcsp/AllInterval-012.c
  testcomp/sv-benchmarks/c/seq-mthreaded/pals_floodmax.3.4.ufo.BOUNDED-6.pals.c
  testcomp/sv-benchmarks/c/hardware-verification-bv/btor2c-lazyMod.needham.4.prop4-back-serstep.c
  testcomp/sv-benchmarks/c/hardware-verification-bv/btor2c-lazyMod.lann.8.prop1-back-serstep.c
  testcomp/sv-benchmarks/c/hardware-verification-bv/btor2c-lazyMod.brp.3.prop1-back-serstep.c
  testcomp/sv-benchmarks/c/hardware-verification-bv/btor2c-lazyMod.adding.2.prop1-back-serstep.c
  testcomp/sv-benchmarks/c/array-fpi/pcompf.c
tasks solved only by tool 2:
  testcomp/sv-benchmarks/c/xcsp/AllInterval-014.c
  testcomp/sv-benchmarks/c/seq-mthreaded/pals_lcr.5.1.ufo.UNBOUNDED.pals.c
  testcomp/sv-benchmarks/c/seq-mthreaded/pals_lcr.5.1.ufo.BOUNDED-10.pals.c
  testcomp/sv-benchmarks/c/reducercommutativity/rangesum.i
  testcomp/sv-benchmarks/c/product-lines/email_spec8_productSimulator.cil.c
  testcomp/sv-benchmarks/c/loop-floats-scientific-comp/loop2-1.c
  testcomp/sv-benchmarks/c/loop-floats-scientific-comp/loop1-2.c
  testcomp/sv-benchmarks/c/hardware-verification-bv/btor2c-lazyMod.needham.4.prop1-back-serstep.c
  testcomp/sv-benchmarks/c/fuzzle-programs/06_fuzzle_50x50_0-cycle.c
```


### Memory usage tool 1

Memory usage: 267159376
Total tests: 1215
Average memory usage: 21988426.01

### Memory usage tool 2
Memory usage: 266837680
Total tests: 1215
Average memory usage: 21961948.97

I only used 4 threads but memory wise it seems to have no impact. So I'm closing this

PR #371, opened 2025-06-12
fix normalization error on floats

felixL-K:
An error was introduced with normalization of floats : 
`x < y` is not equivalent to `not(x >= y)` for floats when x or y is nan

### DIFF
--- a/test/unit/test_expr.ml
+++ b/test/unit/test_expr.ml
@@ -614,6 +614,16 @@ let test_simplify_concat_i32_symbol _ =
   let concated = Expr.concat d (Expr.concat c (Expr.concat b a)) in
   check concated x
 
+let test_fp_nan_not_geffects _ =
+  let open Infix in
+  let ty = Ty.Ty_fp 32 in
+  let x = symbol "x" ty in
+  let y = symbol "y" ty in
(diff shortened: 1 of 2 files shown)


test/unit/test_expr.ml
```diff
@@ -614,13 +614,24 @@ let test_simplify_concat_i32_symbol _ =
...
   check concated x
 
+let test_fp_nan_not_geffects _ =
+  let open Infix in
+  let ty = Ty.Ty_fp 32 in
+  let x = symbol "x" ty in
+  let y = symbol "y" ty in
+  (* x != x makes isNaN(x) *)
+  let expr = Expr.unop ty Not (Expr.relop ty Ge x y) in
+  let expected = Expr.unop ty Not (Expr.relop ty Le y x) in
```
filipeom, line 624:
```suggestion
  let expr = Expr.unop Ty_bool Not (Expr.relop ty Ge x y) in
  let expected = Expr.unop Ty_bool Not (Expr.relop ty Le y x) in
```

The `Not` should be encoded as the boolean operator of negation. Other than that every thing looks good to go!

felixL-K:
All done here.

Issue #363, opened 2025-06-01
Resolve SMT-COMP's QF_S parsing errors


filipeom:
Current status on `QF_S`

smtml-z3:
  Passed: 13744/22172 (61.99%)

PR #372, opened 2025-06-12
Missing RegExp operators

joaomhmpereira:
Adds support for: `re.all`, `re.allchar`, `re.none`, and `re.inter`. Should greatly improve #363.


src/smtml/rewrite.ml
```diff
@@ -50,12 +50,16 @@ let rec rewrite_expr (type_map, expr_map) hte =
...
-      match Symb_map.find_opt sym type_map with
-      | None -> (
-        match Symb_map.find_opt sym expr_map with
-        | None -> Fmt.failwith "Undefined symbol: %a" Symbol.pp sym
-        | Some expr -> expr )
-      | Some ty -> Expr.symbol { sym with ty }
+      match Symbol.name sym with
+      | Simple "re.all" | Simple "re.none" | Simple "re.allchar" ->
+        Expr.app { sym with ty = Ty_regexp } []
+      | _ -> (
```
filipeom, line 56: Sorry to make this request. Your PR is very tidy and well-written! I'm assuming these symbols are under the `Term` namespace, right? If so, could you move this code into the pattern match here:

https://github.com/formalsec/smtml/blob/a6ecea087f6c0e8cd2d1aaa54545c7721a56dd72/src/smtml/smtlib.ml#L67-L75

I'm suggesting this because I did something very similar for the rounding modes, and this case looks quite similar. Also, the type can be inferred at compile time, and this helps avoid duplicating similar logic in two places.

Everything else looks perfect!

joaomhmpereira: I moved it to the pattern match you referred but had to create them as symbols instead of apps to make it work. Is this what you meant?
Also, I need to add a new stub to the cvc5 bindings in order to support the regexp sort. So, won't be available in cvc5 for now

filipeom: > I moved it to the pattern match you referred but had to create them as symbols instead of apps to make it work. Is this what you meant?

Aw ok, but it still works the same being a symbol vs an app right?

> Also, I need to add a new stub to the cvc5 bindings in order to support the regexp sort. So, won't be available in cvc5 for now

I think it should be fine for the time being

joaomhmpereira: > Aw ok, but it still works the same being a symbol vs an app right?
Yes, I think it's the same

> I think it should be fine for the time being
I already have the changes locally and plan to open a PR and publish a new version of the bindings, since cvc5 has already released a newer version (1.2.1).

filipeom:
> Z3's OCaml API does not seem to support the `str.replace_all` operator, only `str.replace` https://z3prover.github.io/api/html/ml/Z3.Seq.html

Weird, but I was able to pass an SMT formula to z3 with `str.replace_all`. Maybe the function is just missing from the API. In any case this might be a problem for another PR. Do you think this is ready to merge?

filipeom:
Thanks!

PR #374, opened 2025-06-20
Replace int2str and str2int with solver operators

joaomhmpereira:
Replaces `int2str` and `str2int` (currently represented as uninterpreted functions) with the actual solver functions for these operations

### DIFF
--- a/src/smtml/bitwuzla_mappings.default.ml
+++ b/src/smtml/bitwuzla_mappings.default.ml
@@ -128,12 +128,6 @@ module Fresh_bitwuzla (B : Bitwuzla_cxx.S) : M = struct
     let to_real _ =
       Fmt.failwith "Bitwuzla_mappings: Int.to_real not implemented"
 
-    let to_string _ =
-      Fmt.failwith "Bitwuzla_mappings: Int.to_string not implemented"
-
-    let of_string _ =
-      Fmt.failwith "Bitwuzla_mappings: Int.of_string not implemented"
-
     let add _ = Fmt.failwith "Bitwuzla_mappings: Int.add not implemented"
 
     let sub _ = Fmt.failwith "Bitwuzla_mappings: Int.sub not implemented"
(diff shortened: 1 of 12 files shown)


filipeom (commented):
All seems good, but can you make the function `to_int` and `of_int` in the `String` module instead of `to_string` and `of_string`? To follow the smtlib standard

joaomhmpereira:
Currently we already have `String_to_int` and `String_from_int` defined in the `String` module ([here](https://github.com/formalsec/smtml/blob/801e1b76c9372c9be14aa8b23724f2b3ecb21473/src/smtml/mappings.ml#L250)).

This PR was a fix for the `OfString` and `ToString` operators used in the `Int` module ([here](https://github.com/formalsec/smtml/blob/801e1b76c9372c9be14aa8b23724f2b3ecb21473/src/smtml/mappings.ml#L154)). Should these be pointing to the functions in the `String` module to avoid duplication?

filipeom:
> Currently we already have `String_to_int` and `String_from_int` defined in the `String` module ([here](https://github.com/formalsec/smtml/blob/801e1b76c9372c9be14aa8b23724f2b3ecb21473/src/smtml/mappings.ml#L250)).

Ah okok.

> This PR was a fix for the `OfString` and `ToString` operators used in the `Int` module ([here](https://github.com/formalsec/smtml/blob/801e1b76c9372c9be14aa8b23724f2b3ecb21473/src/smtml/mappings.ml#L154)). Should these be pointing to the functions in the `String` module to avoid duplication?

Hmm, with this, I still don't like that we can write:

```ocaml
Expr.cvtop Ty_str String_from_int (Int 0) 
(* To mean the same as: *)
Expr.cvtop Ty_int ToString (Int 0)
```

I'd like us to start adopting a [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) philosophy. And since we're following the SMT-LIB standard to some extent in how we name functions and so on, I propose that we no longer accept this: `Expr.cvtop Ty_int ToString (Int 0)`.
The only acceptable way should be: `Expr.cvtop Ty_str String_from_int (Int 0)`, which aligns with how the SMT-LIB standard does it. Likewise for `Expr.cvtop Ty_int OfString`.

So, just:

1. Remove the `ToString` and `OfString` cvtop operators from `Int`
2. I'll make a PR in ECMA-SL to use these operators instead.

filipeom (commented):
Feel free to delete dead code

PR #375, opened 2025-06-21
Fix QF_SLIA parsing errors

joaomhmpereira:
Don't know if this way of propagating types is correct. It does get rid of parsing errors


joaomhmpereira:
Had to do this for the `re.inter` operator:
```ocaml
| "re.inter", [ a; b ] -> Expr.raw_binop Ty_regexp Regexp_inter a b
| "re.inter", [ a; b; c ] ->
        Expr.raw_binop Ty_regexp Regexp_inter a
          (Expr.raw_binop Ty_regexp Regexp_inter b c)
```
Because, despite being considered a binary operator according to the [SMT-LIB](https://smt-lib.org/theories-UnicodeStrings.shtml), some tests in the QF_SLIA benchmarks use it as a ternary operator.

joaomhmpereira:
Currently passing 99.74% of tests from the QF_SLIA benchmarks

PR #378, opened 2025-06-24
More simplifications

filipeom:


src/smtml/expr.ml
```diff
@@ -509,7 +500,8 @@ let rec cvtop theory op hte =
   match (op, view hte) with
   | Ty.Cvtop.String_to_re, _ -> raw_cvtop theory op hte
   | _, Val v -> value (Eval.cvtop theory op v)
-  | String_to_float, Cvtop (Ty_real, ToString, real) -> real
+  | String_to_float, Cvtop (Ty_real, ToString, hte) -> hte
+  | Reinterpret_float, Cvtop (Ty_real, Reinterpret_int, hte) -> hte
```
zapashcanon, line 504: Is this always true? I am now being paranoid about floats haha. I'm wondering if converting a string to a float may not lead to some approximation that would be lost by doing this optimization?

filipeom: > Is this always true? I am now being paranoid about floats haha. I'm wondering if converting a string to a float may not lead to some approximation that would be lost by doing this optimization?

Ah, I thought you were talking about the simplification I'm introducing in this PR. The string to float and float to string are just uninterpreted operators used by ECMA-SL. So it's ok if this isn't 100% precise.

PR #393, opened 2025-07-26
[fix] Construction phormulas with int equality

Kakadu:
I discovered that I can't construct expressions with equality relation. 
I don't know whether it is a bug or not, but this is a fix.

### DIFF
--- a/src/smtml/eval.ml
+++ b/src/smtml/eval.ml
@@ -124,6 +124,7 @@ module Int = struct
       | Le -> ( <= )
       | Gt -> ( > )
       | Ge -> ( >= )
+      | Eq -> Int.equal
       | _ ->
         Fmt.failwith {|relop: Unsupported int operator "%a"|} Ty.Relop.pp op
     in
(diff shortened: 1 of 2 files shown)


filipeom:
Hi! 

Thanks for the PR! 

This is because equality is not defined in the [Reals\_Ints](https://smt-lib.org/theories-[Reals_Ints](https://smt-lib.org/theories-Reals_Ints.shtml).shtml) theory, for example. Instead, it is defined in the [Core](https://smt-lib.org/theories-Core.shtml) theory.

The intended usage was as follows:

```ocaml
(* You specify equality between expressions using `Ty_bool` *)
let works = Expr.relop Ty_bool Eq (Expr.value (Int 0)) (Expr.value (Int 0)) in
(* Instead of using `Ty_int` *)
let doesnt_work = Expr.relop Ty_int Eq (Expr.value (Int 0)) (Expr.value (Int 0)) in
...
```

However, I’ve written `relop Ty_int Eq ...` and similar expressions many times myself, so I think it makes sense to include this patch and allow this usage in other theories as well. (I will work on this in this issue: #394)

Thank you so much!

PR #397, opened 2025-08-11
Add set-info status tracking for .smt2 files

joaomhmpereira:
Add support for tracking (set-info :status …) in .smt2 files.
The interpreter now compares the expected status against the solver’s result and raises an error if it differs (only for sat ↔ unsat mismatches).


src/smtml/interpret.ml
```diff
@@ -69,15 +76,35 @@ module Make (Solver : Solver_intf.S) = struct
...
+        Log.debug (fun k -> k "Unrecognised status value: %a" Expr.pp st);
+        None )
+    | _ -> None
+
+  let extract_status (script : Ast.script) : [ `Sat | `Unsat | `Unknown ] option
+      =
+    List.fold_left
+      (fun acc cmd ->
+        match cmd with Ast.Set_info term -> parse_status term | _ -> acc )
+      None script
```
filipeom, line 96:
```suggestion
    List.find_map
      (fun cmd ->
        match cmd with Ast.Set_info term -> parse_status term | _ -> None )
      script
```

Doesn't need to iterate through all

src/smtml/interpret.ml
```diff
@@ -69,15 +76,35 @@ module Make (Solver : Solver_intf.S) = struct
...
+      match Expr.view st with
+      | Symbol { name = Simple "sat"; _ } -> Some `Sat
+      | Symbol { name = Simple "unsat"; _ } -> Some `Unsat
+      | Symbol { name = Simple "unknown"; _ } -> Some `Unknown
+      | _ ->
+        Log.debug (fun k -> k "Unrecognised status value: %a" Expr.pp st);
+        None )
+    | _ -> None
+
+  let extract_status (script : Ast.script) : [ `Sat | `Unsat | `Unknown ] option
```
filipeom, line 91:
```suggestion
  let find_expected_status (script : Ast.script) : [ `Sat | `Unsat | `Unknown ] option
```

nitpicking

src/smtml/interpret_intf.ml
```diff
@@ -9,6 +9,7 @@ type 'a state =
   { stmts : Ast.script  (** The script being executed *)
   ; smap : (string, Ty.t) Hashtbl.t  (** A mapping of variable names to types *)
   ; solver : 'a  (** The underlying solver instance *)
+  ; status : [ `Sat | `Unsat | `Unknown ] option  (** The status of the execution*)
```
filipeom, line 12:
```suggestion
  ; expected_status : [ `Sat | `Unsat | `Unknown ] option  (** The status of the execution*)
```
Nitpicking

test/cli/smt2/test_smt2.t
```diff
@@ -111,3 +111,11 @@ Test uninterpreted:
   sat
   (model
     (x f32 0.))
+
+Test status tracking:
+  $ smtml run test_set_info_status.smt2
+  sat
+  smtml: internal error, uncaught exception:
+         Failure("Expected status: unsat, but solver returned sat")
```
filipeom, line 119: Can you maybe add a cli option, i.e. `--no-strict-status` that simply logs the error instead of using failwith? This is a breaking change so I want to allow people to have the option to disable this if they want.

src/smtml/interpret.ml
```diff
@@ -26,10 +26,17 @@ module Make (Solver : Solver_intf.S) = struct
...
+      let actual = Solver.check solver assumptions in
+      ( match actual with
       | `Sat -> Fmt.pr "sat@."
       | `Unsat -> Fmt.pr "unsat@."
       | `Unknown -> Fmt.pr "unknown@." );
+      ( match (state.status, actual) with
+      | Some `Sat, `Unsat ->
+        Fmt.failwith "Expected status: sat, but solver returned unsat"
+      | Some `Unsat, `Sat ->
+        Fmt.failwith "Expected status: unsat, but solver returned sat"
```
filipeom, line 38: Maybe something like this for the flag suggestion. You'd pass it through main until this function

```suggestion
        if no_strict_status then Log.error ...
        else  Fmt.failwith "Expected status: unsat, but solver returned sat"
```

PR #432, opened 2025-10-21
Make sure that what is printed in the smtml format is parseable

hra687261:
The purpose of this PR is to make reproducing errors/bugs/issues easier, it is related to https://github.com/OCamlPro/owi/issues/631, and the idea is to make sure that when we print smtml queries (in its AST) we can parse back the same thing and rerun it through smtml. It is currently not the case.

This PR:
- Replaces some `Format`s with `Fmt`s
- Renames `Expr.pp_smt` to `Expr.pp_smtml` (since it prints in the smtml format not smt, the name is now less ambiguous and the doc is updated) 
- Adds a `to-smtml` which simply uses `Expr.pp_smtml` to print the parsed/typed content of a `.smtml` file, I only added this command for testing purposes, I don't think it would be useful for something else (other than potentially formatting but the current output is not particularly well formatted anyway) so I can remove it if you want me to.
- When signed operations are printed "_s" is added to them, but when they are parsed, they are expected to not have the "_s" eg. < over integers is parsed "int.lt" and printed as "int.ls_s". I made the simplest/laziest change which consists in making it possible to parse both the operation with and without "_s" when it is signed and it seems to work. (I am open to better/cleaner solutions) 
- removes commas between printed asserts in the smtml format since they are not used in it.

Can be tested with `dune exec -- smtml to-smtml test/cli/smtml/test_lia.smtml > out.smtml` and then running smtml on `out.smtml`, which works with  bbd38bf and 85121a3 and doesn't without them.

### DIFF
--- a/doc/examples.mld
+++ b/doc/examples.mld
@@ -153,8 +153,8 @@ let solve (toys : toy list) (packs : pack list) (daily_max : int) =
 
   (* Maximise *)
   match Z3.maximize opt profit with
-  | Some v -> Format.printf "%a@." Value.pp v
-  | None -> Format.printf "Unfeasible@."
+  | Some v -> Fmt.pr "%a@." Value.pp v
+  | None -> Fmt.pr "Unfeasible@."
 
 let () =
   let num_toys : int = read_int () in


filipeom (approved):
Thank you so much for this! I've been kinda neglecting the smtml format as we can parse smt-lib. But since we don't have a `Smtml.Expr.t -> smt-lib` yet this is very helpful! Thanks!

PR #436, opened 2025-10-23
Expose solver mappings to declare non-parametric datatypes

filipeom:
This is to allow creating examples such as this: 

```smt2
(declare-datatype Ptr ((mk-ptr (loc Int) (ofs Int))))
(declare-const p1 (Ptr))
(assert (= p1 (mk-ptr 10 8)))
(assert (= (loc p1) 10))
(assert (= (ofs p1) 8))
(check-sat)
(get-model)
```

cc @hra687261 (and anyone who wants to help :sweat_smile:). I'm not very comfortable with ADTs in SMT, so I would appreciate your help with this one. I left some of my questions below:

 Do you thing the API for `Adt` and `Adt.Cons` are ok? In particular, do you think the `get_constructor`, `get_recognizers`, and `get_accessors` provide an awkward api? Maybe having `val get_constructor : t -> cons_name:string -> Cons.t` would be better? (Same logic for the others)
   
I like the `Dolmen_std.Expr.Term.define_adt` interface but I don't really understand how the API works. Do you have some example usages so I can finish the API for dolmen to have adt support in Alt-Ergo?

### DIFF
--- a/.ocamlformat
+++ b/.ocamlformat
@@ -1,4 +1,4 @@
-version=0.27.0
+version=0.28.1
 assignment-operator=end-line
 break-cases=fit
 break-fun-decl=wrap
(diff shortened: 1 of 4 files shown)


filipeom:
Thanks for the help @hra687261. If we find this API is not very good we can improve on in the future

PR #449, opened 2025-11-18
put the cache info (hits/misses) in solver stats

S41d:
As discussed, this puts the cache hits and misses inside solver stats.


filipeom (approved):
Thanks for taking care of the test failures!

PR #451, opened 2025-11-18
add the ability to inline a symbol in a set

redianthus:
This will be used in Owi (or maybe the simplification makes sense directly in Smt.ml, we'll see).

The idea is that, when we add something like `symbol_0 = 42` in the PC. Then, we can propagate the equality and replace `symbol_0` by `42` in order to simplify the PC and save time/solver calls in the remaining execution.

This is a quick draft, I'll finish it soon.

### DIFF
--- a/src/smtml/expr.ml
+++ b/src/smtml/expr.ml
@@ -115,19 +115,21 @@ let rec ty (hte : t) : Ty.t =
   | Symbol x -> Symbol.type_of x
   | List _ -> Ty_list
   | App (sym, _) -> begin match sym.ty with Ty_none -> Ty_app | ty -> ty end
-  | Unop (ty, _, _) -> ty
-  | Binop (ty, _, _, _) -> ty
   | Triop (_, Ite, _, hte1, hte2) ->
     let ty1 = ty hte1 in
-    let ty2 = ty hte2 in
-    assert (Ty.equal ty1 ty2);
+    assert (
+      let ty2 = ty hte2 in
+      Ty.equal ty1 ty2 );
     ty1
-  | Triop (ty, _, _, _, _) -> ty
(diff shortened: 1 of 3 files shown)


src/smtml/expr.ml
```diff
@@ -138,22 +140,65 @@ let rec ty (hte : t) : Ty.t =
...
+      let v2 = aux v2 in
+      let v3 = aux v3 in
+      make @@ Triop (ty, op, v1, v2, v3)
+    | Cvtop (ty, op, v) ->
+      let v = aux v in
+      make @@ Cvtop (ty, op, v)
+    | Relop (ty, op, v1, v2) ->
+      let v1 = aux v1 in
+      let v2 = aux v2 in
+      make @@ Relop (ty, op, v1, v2)
```
filipeom: Say you have `x = 42`, won't this build the expression `42 = 42`? Maybe we should use smart constructors? That way we get `True`, or is this ok for what you're doing?

filipeom (approved):
> @filipeom, should be good to review now. It finally decided that the function should take an expression and a map from symbol to expression as input, which is how we'll be using it in Owi. Let me know if you can think of a better API (or naming :-)).
> 
> While reading the code around, I also noticed some pattern matching that could be simplified a little bit and decided to commit all of this together, sorry about this 😅

Thanks for taking the time to do this! I had a small question below. Regarding naming, I'm probably the worst person to suggest any change :sweat_smile: I actually like the API you're proposing

redianthus:
@filipeom PTAL :)

filipeom (approved):
LGTM! I can merge if you want :smile:

filipeom:
Thanks! 🙏

PR #452, opened 2025-11-19
Feature: Random_seed parameter

tiferrei:
Hi there, 

Thanks a lot for this project! I just added a parameter for fixing the random seed (for Z3), and thought I'd submit a PR to upstream if anyone else is interested.

I had to remove the `%%VERSION%%` template from `main.ml`, as it was giving me compilation issues. AFAIK this seems to not be populated at any point, but I am happy to remove that commit.

Thanks,

Tiago

### DIFF
--- a/src/smtml/cvc5_mappings.default.ml
+++ b/src/smtml/cvc5_mappings.default.ml
@@ -519,7 +519,8 @@ module Fresh_cvc5 () = struct
       | Unsat_core ->
         Solver.set_option slv "produce-unsat-cores" (string_of_bool v)
       | Ematching -> Solver.set_option slv "e-matching" (string_of_bool v)
-      | Parallel | Num_threads | Debug | Random_seed -> ()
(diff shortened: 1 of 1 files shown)


filipeom (approved):
Thank you so much your contribution! Everything looks good to me. Let me know if you have any issues using smtml!

tiferrei:
Thanks a lot for getting this merged so quickly! 

Now that you mention it, I have had some issues getting the `cvc5` lib installed through nix (`nix build 'github:tweag/opam-nix#cvc5.latest'`), probably something to do with the static libraries and macOS, however this seems to be a Nix specific problem.

Additionally, I noticed that there is no support for tuples or enums, which at first gave me some issues but I was able to work around with lists and some hacky integer representations. Thanks again!

tiferrei:
> For enums, we already have the mappings for them in smtml 

Ah very nice! No worries, it is not a key feature for me

PR #461, opened 2025-11-24
Add the names of SMT solvers to their mappings (and log them with SMT queries)

hra687261:
Adding them as hard coded strings to the mappings seemed to be simplest way, but their might be more elegant ways to do it. (by getting it from the command line options maybe?)


filipeom:
> Adding them as hard coded strings to the mappings seemed to be simplest way, but their might be more elegant ways to do it. (by getting it from the command line options maybe?)

I feel it's ok to leave them as static strings in the program. Getting them from the command line would be tricky for people who just want to use the library. Also, then the name would also have to be an `string option` which might be confusing?

PR #479, opened 2025-12-10
Add a was_interrupted ref that indicates if Solver.interrupt was call…

hra687261:
…ed on a solver or not

Edit: not sure if this is necessary actually (cf: https://github.com/OCamlPro/owi/pull/861#discussion_r2606787117)


hra687261:
> So feel free to merge this if you end up needing it.

Thanks!

PR #480, opened 2025-12-11
Fix incorrect odoc documentation syntax 

filipeom:


filipeom:
Haha no worries! I also didn't catch it :sweat_smile:

PR #482, opened 2025-12-12
use Set.Make instead of PatriciaTree.MakeHashconsedSet

### LINKED ISSUE #481: crash coming from PatriaTree
Hi @filipeom,

This one took me a long time to figure out. I finally managed to get a backtrace for a bug we had for a long time and here is what I got:

```shell-session
$ dune exec -- owi c doc/src/sap/maze.c --model-out-file=out.scfg -w2
...
owi: [ERROR] a worker ended with exception Invalid_argument("index out of bounds"), backtrace is: 
             Raised by primitive operation at Stdlib__Weak.Make.resize.add_weak in file "weak.ml", line 223, characters 16-23
Called from Stdlib__Weak.Make.iter_weak.iter_bucket in file "weak.ml", line 166, characters 16-34
Called from Stdlib__Array.iteri in file "array.ml", line 157, characters 31-51
Called from Stdlib__Weak.Make.resize in file "weak.ml", line 226, characters 6-26
Called from Stdlib__Weak.Make.merge.(fun) in file "weak.ml", line 298, characters 40-66
Called from PatriciaTree__Nodes.HashconsedSetNode.try_find in file "src/nodes.ml", line 347, characters 12-42
Called from PatriciaTree__Functors.MakeCustomHeterogeneousMap.insert_for_union.loop in file "src/functors.ml", line 545, characters 61-73
Called from PatriciaTree__Functors.MakeCustomHeterogeneousMap.insert_for_union.loop in file "src/functors.ml", line 545, characters 61-73
Called from PatriciaTree__Functors.MakeCustomHeterogeneousMap.insert_for_union.loop in file "src/functors.ml", line 544, characters 54-66
Called from PatriciaTree__Functors.MakeCustomHeterogeneousMap.insert_for_union in file "src/functors.ml", line 547, characters 9-15
Called from Owi__Union_find.Make.add.(fun) in file "src/data_structures/union_find.ml", line 122, characters 44-72
Called from Stdlib__Map.Make.update in file "map.ml", line 287, characters 18-28
Called from Stdlib__Map.Make.update in file "map.ml", line 292, characters 21-33
Called from Stdlib__Map.Make.update in file "map.ml", line 295, characters 21-33
Called from Owi__Union_find.Make.add in file "src/data_structures/union_find.ml", lines 114-125, characters 6-28
Called from Owi__Symbolic_path_condition.add_one in file "src/symbolic/symbolic_path_condition.ml", line 17, characters 6-56
Called from Owi__Thread.Make.add_pc in file "src/symbolic/thread.ml", line 75, characters 13-47
Called from Owi__Symbolic_choice.Make.add_pc in file "src/symbolic/symbolic_choice.ml", line 368, characters 23-45
Called from Owi__Symbolic_choice.CoreImpl.State.bind.(fun) in file "src/symbolic/symbolic_choice.ml", line 146, characters 14-19
Called from Owi__Symbolic_choice.CoreImpl.Schedulable.bind.(fun).unfold_status in file "src/symbolic/symbolic_choice.ml", line 39, characters 27-32
Called from Owi__Symbolic_choice.CoreImpl.Schedulable.run in file "src/symbolic/symbolic_choice.ml" (inlined), line 28, characters 6-13
Called from Owi__Symbolic_choice.CoreImpl.Schedulable.bind.(fun) in file "src/symbolic/symbolic_choice.ml", line 47, characters 24-36
Called from Owi__Symbolic_choice.CoreImpl.Schedulable.run in file "src/symbolic/symbolic_choice.ml" (inlined), line 28, characters 6-13
Called from Owi__Symbolic_choice.CoreImpl.Schedulable.choose.(fun) in file "src/symbolic/symbolic_choice.ml", line 59, characters 47-56
Called from Owi__Symbolic_choice.CoreImpl.Schedulable.run in file "src/symbolic/symbolic_choice.ml" (inlined), line 28, characters 6-13
Called from Owi__Symbolic_choice.CoreImpl.Schedulable.bind.(fun) in file "src/symbolic/symbolic_choice.ml", line 47, characters 24-36
Called from Owi__Symbolic_choice.CoreImpl.Schedulable.run in file "src/symbolic/symbolic_choice.ml" (inlined), line 28, characters 6-13
Called from Owi__Symbolic_choice.CoreImpl.Schedulable.bind.(fun) in file "src/symbolic/symbolic_choice.ml", line 47, characters 24-36
Called from Owi__Symbolic_choice.CoreImpl.Schedulable.run in file "src/symbolic/symbolic_choice.ml" (inlined), line 28, characters 6-13
Called from Owi__Symbolic_choice.CoreImpl.Schedulable.bind.(fun) in file "src/symbolic/symbolic_choice.ml", line 47, characters 24-36
Called from Owi__Symbolic_choice.CoreImpl.Schedulable.run in file "src/symbolic/symbolic_choice.ml" (inlined), line 28, characters 6-13
Called from Owi__Symbolic_choice.CoreImpl.Schedulable.bind.(fun) in file "src/symbolic/symbolic_choice.ml", line 47, characters 24-36
Called from Owi__Symbolic_choice.CoreImpl.Schedulable.run in file "src/symbolic/symbolic_choice.ml" (inlined), line 28, characters 6-13
Called from Owi__Symbolic_choice.CoreImpl.Scheduler.work.(fun) in file "src/symbolic/symbolic_choice.ml", line 92, characters 43-66
Called from Stdlib__Fun.protect in file "fun.ml", line 34, characters 8-15
Re-raised at Stdlib__Fun.protect in file "fun.ml", line 39, characters 6-52
Called from Synchronizer.work_while in file "src/synchronizer.ml", lines 64-66, characters 4-42
Called from Owi__Symbolic_choice.CoreImpl.Scheduler.spawn_worker.(fun) in file "src/symbolic/symbolic_choice.ml", lines 103-105, characters 14-64
```

To me it seems that this library is not thread-safe (and is using `Obj` a lot, which seems risky). I tried to replace it with `Set` and the error was no longer reproducible.

redianthus:
Fix #481.

I will also open an issue upstream.

### DIFF
--- a/src/smtml/expr.ml
+++ b/src/smtml/expr.ml
@@ -830,6 +830,8 @@ module Set = struct
 
   let hash = Hashtbl.hash
 
+  let to_list s = elements s
+
   let pp fmt v =
     Fmt.pf fmt "@[<hov 1>%a@]"
       (Fmt.iter iter ~sep:(fun fmt () -> Fmt.pf fmt "@;") pp)


redianthus:
Yes, I think there are a lot of questions about thread-safety that we'll have to address at some point. The one described https://github.com/formalsec/smtml/issues/481 was outstanding in the sense that it would either make Owi crash, miss a bug, or make it enter a deadlock (that was before I fixed Owi and Synchronizer, now it would print the error properly).

Whereas all the issues you are mentioning do not seem to be causing issues yet, from my various testing at least. My current guess is that they could indeed lead to safety issues, but the probabilities seem to be small enough that it does not show up. That being said, it is very likely that even if they don't compromise safety (on my current testing at least, I wouldn't be surprised if they ever explode in our face one day), they may be silently compromising performances (for instance, the same values being created in two different threads ending up as 
two different values in the hash-consing table because they were created almost at the same time).

Here are my thinking about each of them.

Regarding the current PR, I propose that we merge it as is (I still need to fix 4.14). And I'll open an issue so that we think about replacing `Set` by something more efficient but thread-safe in the future.

> I think expression creation should be thread-safe as we want to reduce the memory footprint of programs. For that reason, we should probably change the hashconsing of expressions to use:

Indeed, and when I saw the crash we are talking about, look at what was my first try : https://github.com/formalsec/smtml/compare/main...redianthus:smtml:hc_strong#diff-0a384bcf470e710dea43d9c3ac543c07b680517e2a0b9ccfef6ac5d9139bf8b4R89 :sweat_smile: 

The thing is that I'm afraid this one is going to make performance worst a lot. I'd like to benchmark it first before we start using it on Smt.ml. If the performances are fine, then I don't mind using it directly. Otherwise, I have some ideas on how we could improve the situation (one hashcons cache per type of expression, or simply use a lock-free hash table). I'll open an issue about this one too.

> The issue is that we're using Cache in simplify to memoize simplified terms:

I forgot about this one, which indeed is more annoying to solve. We could try to make it thread-safe too but then, if we go with a synchronization approach, the cost may end up higher (and not only for mono-core users which are going to pay more anyway, but even for multi-core ones!) than the cost of "simplifying many times" - in which case we'd better keep it as is and acknowledge it is a "best effort" cache. Also, I am not sure how much this cache it useful, it would be good to measure it (if it is not, we can simply remove it and the problem will be solved).

PR #485, opened 2025-12-14
Add function `Value.default_of_type` (#483)

### LINKED ISSUE #483: provide a dummy value for a given type
Hi,

I have some code where I have a set of symbols. For some of them, I have a model, but for the others I don't because they are actually useless. Still, I'd like to display a value for all of them.

I can do it by having something like:

```ocaml
      match ty with
      | Smtml.Ty.Ty_bool -> Smtml.Value.of_string ty "false"
      | Ty_bitv n ->
        Ok (Smtml.Value.Bitv (Smtml.Bitvector.make (Z.of_string "0") n))
      | Ty_fp 32 -> Smtml.Value.of_string ty "0.0"
      | Ty_fp 64 -> Smtml.Value.of_string ty "0.0"
      | _ty ->
        (* if this fails, it may be that we changed something and a new type should be added here *)
        assert false
    in
```

But I though it would be better in Smtml as a "Smtml.Value.dummy_of_type` or something. What do you think?

filipeom:
Closes #483

### DIFF
--- a/src/smtml/value.ml
+++ b/src/smtml/value.ml
@@ -83,6 +83,20 @@ let map v f = match v with Nothing -> Nothing | _ -> f v
 
 let ( let+ ) = map
 
+let default_of_type = function
+  | Ty.Ty_bool -> False
+  | Ty_int -> Int 0
+  | Ty_real -> Real 0.0
+  | Ty_str -> Str ""
+  | Ty_bitv m -> Bitv (Bitvector.make Z.zero m)
+  | Ty_fp 32 -> Num (F32 0l)
+  | Ty_fp 64 -> Num (F64 0L)
+  | Ty_list -> List []
+  | Ty_unit -> Unit
+  | Ty_none -> Nothing
+  | (Ty_fp _ | Ty_app | Ty_regexp | Ty_roundingMode) as ty ->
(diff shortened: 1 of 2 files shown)


src/smtml/value.ml
```diff
@@ -83,6 +83,20 @@ let map v f = match v with Nothing -> Nothing | _ -> f v
...
+  | Ty_int -> Int 0
+  | Ty_real -> Real 0.0
+  | Ty_str -> Str ""
+  | Ty_bitv m -> Bitv (Bitvector.make Z.zero m)
+  | Ty_fp 32 -> Num (F32 0l)
+  | Ty_fp 64 -> Num (F64 0L)
+  | Ty_list -> List []
+  | Ty_unit -> Unit
+  | Ty_none -> Nothing
+  | (Ty_fp _ | Ty_app | Ty_regexp | Ty_roundingMode) as ty ->
```
redianthus: I am surprised to see `Ty_fp` here @filipeom, is it expected?

filipeom: Atm we only support `Ty_fp 32` and `Ty_fp 64` so that is why the remaining ones fall here. At some point we will support arbitrary sized FPAs.

PR #496, opened 2025-12-22
Query logging: separate check_sat and model producing queries

hra687261:


filipeom (approved):
I wonder if in the future we could to this with spans/logs using open telemetry https://github.com/imandra-ai/ocaml-opentelemetry. 

It would be cool since smtml could provide observability to users who already have an open telemetry collector configured. However, I'm not to sure about serializing the queries to strings :grimacing: I may try this in the future tho.

Thanks for the PR!

PR #504, opened 2026-01-05
Reduce allocations in hashconsing hash function

filipeom:
Seems to improve performance a bit. Not sure if it is just by not using tuples or because of the hash functions of the other modules.

### DIFF
--- a/src/smtml/bitvector.ml
+++ b/src/smtml/bitvector.ml
@@ -25,6 +25,11 @@ let of_int32 v = make (Z.of_int32 v) 32
 
 let of_int64 v = make (Z.of_int64 v) 64
 
+(* Optimized mixer (DJB2 variant). Inlines to simple arithmetic. *)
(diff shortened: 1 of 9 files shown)


filipeom:
> Out of curiosity, is there a simple way to measure the effect on performance of "small changes"?

At the moment we don't have enough observability to do so. 

What I'm doing in our symbolic execution engine for javascript is roughly looking at execution time and also the flamegraph traces of the execution. But for programs that I know suffer from hashconsing issues. 

For example, this PR improves the following javascript program execution time by 3 seconds. Not a big gain, but it is something :sweat_smile: 

```js
var i = 0;
var limit = 5000;
var res = 0;
while (i < limit) {
  res += i;
  i += 1;
}
console.log(res);
``` 

In the future I want to add some tracing to smtml. That way it's easier to measure the effects of small changes. But right now I'll try to pick the low hanging fruit first by looking at flamegraph :sweat_smile:

PR #505, opened 2026-01-05
Fix logging of queries after model generation

hra687261:
Log both the assumptions added before the last `check-sat`, as well as those added with it (through `check-sat-assuming`)


redianthus (approved):

filipeom (approved):
:pray:

PR #513, opened 2026-01-19
Split type `t` into `Imm` values and `Sym` hash-consed nodes

filipeom:
This commit fundamentally changes the representation of expressions (`t`) by lifting constant values out of the hash-consing machinery.

Changes:

- `t` is now a variant: `Imm of Value.t | Sym of expr Hc.hash_consed`.
- Removed `Val` and `Loc` constructors from the underlying `expr` type.
- Constants are no longer hash-consed; they are stored directly as `Imm`.

Potential Performance Improvements:

- Creating constant expressions  no longer requires a hash-cons table lookup or allocation of a hash-cons node.
- The global hash-cons table is no longer polluted with ephemeral constant values.

BREAKING CHANGE: Previously, because all expressions were hash-consed, structural equality implied physical equality (`phys_equal` was sufficient for all comparisons).

This invariant no longer holds for `Imm` values.


redianthus:
What is the impact on performances for ECMA-SL on your hash-consing heavy program you mentioned? (The one where switching from weak to strong reduced the overhead from 40% to 20%)

redianthus:
Yes, there is nothing you should do to get it working. Simply add:

```dune
 (instrumentation
  (backend landmarks --auto)))
```

to your library/executable in dune.

Then :

```shell-session
OCAML_LANDMARKS=on dune exec --instrument-with landmarks --profile release -- owi run test/run/binary_loop.wasm
```

PR #519, opened 2026-01-21
Normalize `not (a || b)` -> `not a && not b` (#517)

### LINKED ISSUE #517: missing canonization
I think we should simplify `not(a || b)` to `not(a) and not(b)`, it should allow more simplifications in the simplify recursive function but should also improve a lot path condition slicing! WDYT @filipeom ?

filipeom:
Closes #517


filipeom:
> I am never sure about it so I'll ask: does it take care of intermediate int/bool conversion? I.e. if there is a few conversions between the various operators, does it see through them?

I don't know if I follow. Can you give me an example?

redianthus:
Should we also add this one to smtml: https://github.com/OCamlPro/owi/blob/ae97087ff172309d26acc468ff2e2bf9cbf299a4/src/symbolic/symbolic_i32.ml#L32 ? (the converse)

PR #521, opened 2026-01-21
Add zero-cost typed interface using phantom types

filipeom:
Introduces a Typed wrapper module to provide compile-time safety for SMT types (Int, Real, BitVec) without altering the runtime AST. This includes a `Make_Bitv` functor to generalize bitvector widths like `Bitv32` and `Bitv64`.

@redianthus This approach was chosen over a GADT-based deep embedding to avoid breaking changes in the API as well as the performance penalties of "boxed" existential types.
GADTs would require wrapping terms to store them in heterogeneous structures complicating efficient hash-consing. Phantom types ensure safety with zero runtime overhead.

### DIFF
--- a/src/smtml/typed.mli
+++ b/src/smtml/typed.mli
@@ -1,5 +1,7 @@
+(** The type of a type witness. *)
 type 'a ty = private Ty.t
 
+(** The type of a typed expression (term). *)
 type 'a expr = private Expr.t
 
 type real
@@ -19,19 +21,33 @@ type float32
 type float64
 
 module Unsafe : sig
+  (** [wrap e] promotes a raw untyped expression [e] to a typed expression.
+      {b Warning.} This is unsafe because it bypasses the type checker. *)
   val wrap : Expr.t -> 'a expr
 
+  (** [unwrap e] extracts the raw untyped expression from a typed expression
+      [e]. *)
   val unwrap : 'a expr -> Expr.t
 end
 
+(** [view e] typed alias for {!Expr.view}. Allows retrieving the underlying
+    structure of the expression [e]. *)
 val view : 'a expr -> Expr.expr
 
+(** [simplify e] typed alias for {!Expr.simplify}. Performs algebraic
+    simplifications on the expression [e]. *)
 val simplify : 'a expr -> 'a expr
 
+(** [symbol ty name] creates a symbolic constant of type [ty] with the given
+    [name]. *)
 val symbol : 'a ty -> string -> 'a expr
 
+(** [get_symbols es] returns a list of all unique symbols appearing in the list
+    of expressions [es]. *)
 val get_symbols : 'a expr list -> Symbol.t list
 
+(** [ptr value base] creates a 32-bit pointer expression by adding [value] to
+    the [base] address. *)
 val ptr : int32 -> bitv32 expr -> bitv32 expr
 
 module Types : sig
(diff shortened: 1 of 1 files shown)


redianthus:
It seems like a good idea! Indeed, satisfying type safety from the outside (i.e. the exposed API) is probably enough and makes things much simpler. I like it!

redianthus:
I started trying to port Owi to this. Comments :

- [ ] should we completely hide the untyped interface of `Expr` to prevent errors on the user side?
- [ ] add `val simplify : 'a Typed.t -> 'a Typed.t`?
- [ ] add `val view`?
- [ ] the order of arguments of `symbol` is not the same as `Symbol.make` (type and name inverted), maybe this should be fixed? (unless there's a good reason to put them this way)
- [ ] why is there an `eq` specialized for floats but not other types?
- [ ] providing instanciation of the bv functor for common types is probably useful? (32 and 64 for instance)
- [ ] what about replacing `type 'a t = Expr.t` by `type 'a t = private Expr.t`?
- [ ] `val not_` could be `val not`, right? (only `or_` and `and_` have to be like this AFAIK)
- [ ] I'd put all the boolean functions in their own sub module
- [ ] `pp` functions are missing
- [ ] Float32.ceil is missing
- [ ] Float32.floor is missing
- [ ] Float32.trunc is missing
- [ ] Float32.nearest is missing
- [ ] Float32.copysign is missing
- [ ] Float32.ne is missing
- [ ] many conversion operations on floats are missing, I found no equivalent for :

```ocaml
let convert_i32_s x = cvtop ty ConvertSI32 x

let convert_i32_u x = cvtop ty ConvertUI32 x

let convert_i64_s x = cvtop ty ConvertSI64 x

let convert_i64_u x = cvtop ty ConvertUI64 x

let demote_f64 x = cvtop ty DemoteF64 x

let reinterpret_i32 x = cvtop ty Reinterpret_int x

let of_bits x = cvtop ty Reinterpret_int x

let to_bits x = cvtop (Ty_bitv 32) Reinterpret_float x
```

- [ ] ditto for `Float64` :sweat_smile:

filipeom:
> So, I had a quick look at the simplifications failures in Owi.
> 
> We call `Smtml.Typed.simplify` on:
> 
> ```ocaml
> owi: [ERROR] simplifying expression: (bool.and
>                                       (bool.or (bool.ne symbol_0 0)
>                                        (bool.ne symbol_1 0)) true)
> ```
> 
> But it does not get simplified, because the `e1 || true ---> true` simplification used to be done somewhere else (I guess in some smart constructor that is not being called anymore?), whereas `Smtml.Typed.simplify` calls `Expr.simplify` which seems to only be trying to simplify `e1` and `true` and then rebuilds a conjunction from them.

The generated expression looks a lot like the simplification we added in: https://github.com/formalsec/smtml/pull/519 

I'll investigate further tomorrow

filipeom:
> I tried without the typed inference stuff and with https://github.com/formalsec/smtml/pull/519 and it works, so I guess the issue is somewhere else. 😅

ahh okok. That explains it! I guess these simplifications:
https://github.com/formalsec/smtml/blob/c33fa749c00bc32bd12d5e04772f687594d059d2/src/smtml/expr.ml#L690-L702
are missing in and_ and or_. I pushed a temporary fix: https://github.com/formalsec/smtml/compare/74910e3744bcee2ba679415d5b0f9f5a2e8025bb..27dd74685e604359e699bbe642a0bf1dafe381db which should resolve those issues you were having. Once #348 is merged I can remove `Expr.Bool` and it should also work fine :smile:

PR #530, opened 2026-01-31
remove all > operators and enforce the use of < instead

redianthus:
This is the companion PR of https://github.com/OCamlPro/owi/pull/908 ; it makes normalizing easier and more efficient.

There are three remaining issues :

- tests working on `.smtml` files are not working anymore, I suggest we do not allow using `>` in `.smtml` format;
- tests working on `.smt2` files are not working anymore, this one is more annoying, but I believe it could be fixed by changing a little bit the way the parser works and simply translating `>` into `<` during parsing;
- this is a breaking change, we can either decide it's fine, or instead, keep providing all the `val {gt,ge,gt_u,ge_u}` functions but make sure they simply use the right `{LT,LE,LT_U,LE_U}` constructors. (I'm rather against adding them, but I wouldn't mind doing so)

WDYT?

### DIFF
--- a/README.md
+++ b/README.md
@@ -186,8 +186,8 @@ val cond : Expr.t = (bool.and a (bool.not b))
     let two = Typed.Int.v 2 in
     let x = Typed.symbol Typed.Types.int "x" in
     let y = Typed.symbol Typed.Types.string "y" in
-    Typed.Int.ge (Typed.Int.add x two) y
-Line 5, characters 40-41:
+    Typed.Int.le y (Typed.Int.add x two)
+Line 5, characters 18-19:
 Error: The value y has type string Typed.expr
        but an expression was expected of type Typed.Int.t = int Typed.expr
        Type string is not compatible with type int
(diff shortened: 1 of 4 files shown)


redianthus:
Yes, I think we have to keep it in the smtlib parsing at least, but this is a little bit of work (we can't simply replace a > b by a < b directly in the parser because of the way it is currently written). But maybe for the `.smtml` format we don't care and that would be less work to do :sweat_smile: 

For the API, I'm not sure. It may be more convenient for users, but then, we are simply doing useless work (`let gt x y = lt y x`) and maintaining useless code. I think both options are OK and I don't really mind keeping them in the API.

src/smtml/smtlib.ml
```diff
@@ -274,8 +268,6 @@ module Term = struct
       | "fp.max", [ a; b ] -> Expr.raw_binop Ty_none Max a b
       | "fp.leq", [ a; b ] -> Expr.raw_relop Ty_none Le a b
       | "fp.lt", [ a; b ] -> Expr.raw_relop Ty_none Lt a b
-      | "fp.geq", [ a; b ] -> Expr.raw_relop Ty_none Ge a b
-      | "fp.gt", [ a; b ] -> Expr.raw_relop Ty_none Gt a b
```
filipeom: And these

PR #542, opened 2026-02-16
Add SMTZilla training scripts

hra687261:
This PR is rebased on #540 and should be merged after it. I created a separate PR to make reviewing easier.
This needs a bit more testing, which I will do later and then mark it as ready for review.
I notably don't know if using `dune-site` for this is overkill, but afaik that's the right/safe way to do it.

Note: The python code is a reimplementation of code originally written by @felixL-K

### DIFF
--- a/test/cli/smt2/test_smt2.t
+++ b/test/cli/smt2/test_smt2.t
@@ -114,10 +114,10 @@ Test status tracking:
          Called from Smtml__Interpret.Make.eval in file "src/smtml/interpret.ml", lines 65-66, characters 12-46
          Called from Smtml__Interpret.Make.loop in file "src/smtml/interpret.ml", line 110, characters 8-65
          Called from Dune__exe__Cmd_run.run.run_file in file "src/bin/cmd_run.ml", line 53, characters 13-78
-         Called from Stdlib__List.fold_left in file "list.ml", line 125, characters 24-34
+         Called from Stdlib__List.fold_left in file "list.ml", line 123, characters 24-34
          Called from Dune__exe__Cmd_run.run in file "src/bin/cmd_run.ml", line 99, characters 14-54
-         Called from Cmdliner_term.app.(fun) in file "cmdliner_term.ml", line 24, characters 19-24
-         Called from Cmdliner_eval.run_parser in file "cmdliner_eval.ml", line 35, characters 37-44
+         Called from Cmdliner_term.app.(fun) in file "cmdliner_term.ml", line 22, characters 19-24
+         Called from Cmdliner_eval.run_parser in file "cmdliner_eval.ml", line 41, characters 7-16
   [125]
 
   $ smtml run --no-strict-status test_set_info_status.smt2


dune-project
```diff
@@ -111,3 +114,7 @@
   (re :with-dev-setup)
   (sexplib :with-dev-setup)
   (tls-lwt :with-dev-setup)))
+
+(package
+ (name smtzilla_utils)
+ (sites (share data)))
```
filipeom: Why we put the `dune-site` dependency in smtml, but then create a new package to host data? I say we move `(sites (share data))` to the smtml package.

src/bin/cmd_smtzilla.ml
```diff
@@ -0,0 +1,235 @@
...
+      if exists then
+        Fpath.of_string reqfile_path >>= fun f ->
+        Bos.OS.File.with_ic f
+          (fun ic () -> Fmt.pr "%s" (In_channel.input_all ic))
+          ()
+      else
+        Error
+          (`Msg
+             (Fmt.str "The python requirements file does not exist in: %s"
+                reqfile_path ) )
```
redianthus: FYI, there is something like `Fmt.error_msg` that does exactly ```Error(`Msg(Fmt.str "..."))``` and is quite handy

src/bin/cmd_smtzilla.ml
```diff
@@ -0,0 +1,235 @@
...
+    | Some f -> "--export" :: Fpath.to_string f :: args
+    | None -> args
+  in
+  let args = if run_simulation then "--simulation" :: args else args in
+  let args = if pp_stats then "--pp-stats" :: args else args in
+  let args =
+    if gradient_boost then "--gradient-boost" :: args
+    else "--no-gradient-boost" :: args
+  in
+  let args = if debug then "--debug" :: args else args in
```
redianthus: If think you could directly use the nice syntax which is like:

```ocaml
Bos.Cmd.(v "python3" % one_arg %% many_args % one_arg)
```

You can do `Bos.Cmd.empty` to aovid having to do concatenation by hand everywhere, for instance:

```ocaml
... %% if debug then v "--debug" else empty %% ...
```

Sometimes people also like to do:

```ocaml
let debug = if debug then c "--debug" else Bos.Cmd.empty in

...

Bos.Cmd.(v "python3" ... %% debug %% ...)
```

redianthus:
> Yes, I am not sure what, but I think we had something of the kind in owi, where backtraces were turned on somehow even if we do not do it explicitly (cc @redianthus ), but it's probably not related to this one, because AFAIK this did not happen before this PR, right?

Yes, still an issue I think: https://github.com/OCamlPro/owi/issues/455

src/bin/cmd_smtzilla.ml
```diff
@@ -0,0 +1,235 @@
...
+
+let python_script_path () =
+  let python_script_path =
+    String.concat "/" [ smtzilla_data_dirpath (); __SCRIPT_NAME__ ]
+  in
+  let res =
+    Fpath.of_string python_script_path >>= Bos.OS.File.exists >>= fun exists ->
+    if exists then Ok python_script_path
+    else
+      Error
```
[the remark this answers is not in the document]
hra687261: Done.

src/smtzilla_utils/smtzilla.py
```diff
@@ -0,0 +1,281 @@
...
+        debug_no_gb(data, models, feature_cols)
+
+if args.export:
+    models_json = export_models_to_json(models, feature_cols)
+    print(f"Exporting to: {args.export}")
+    with open(args.export, "w") as f:
+        json.dump(models_json, f)
+
+if args.simulation:
+    simulation(data, models, feature_cols)
```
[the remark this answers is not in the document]
hra687261: Neither am I tbh, this was essentially a rewriting of @felixL-K code into a more of a "functional" way (and with command line options).
So a review from someone more familiar with python is definitely welcome.

I was planning on doing another pass after merging, not on the python code per se, but specifically on how the models are trained/generated, because there are some thresholds in the trees that make some branches of the decision tree unreachable (when you have a negative threshold for example), and I am not how/why we get these, if that is normal or if we simply need to "cleanup" the decision trees.

redianthus:
> It's an example of a marshalled file of queries (but it should've been renamed ...).

OK, so, maybe we can have a smaller one if it's only for illustration purpose? (16MB of marshaled data is a lot!)

Maybe we could avoid having a file at all and show how to generate it in the documentation ?

src/bin/cmd_smtzilla.ml
```diff
@@ -0,0 +1,227 @@
...
+let perdictor_conv =
+  let parse s =
+    match String.lowercase_ascii s with
+    | "gradient-boost" | "gb" -> Ok GradientBoost
+    | "decision-tree" | "dt" -> Ok DecisionTree
+    | _ -> Error (`Msg (Printf.sprintf "Unknown model type: %s" s))
+  in
+  let print ppf = function
+    | GradientBoost -> Format.fprintf ppf "GradientBoost"
+    | DecisionTree -> Format.fprintf ppf "DecisionTree"
```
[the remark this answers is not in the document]
hra687261: Indeed, forgot to change it in `print`, done.

src/bin/cmd_smtzilla.ml
```diff
@@ -0,0 +1,208 @@
...
+let script_name = R.failwith_error_msg (Fpath.of_string "smtzilla.py")
+
+let smtzilla_data_dirpath () =
+  match Smtml_sites.Sites.data with
+  | [ dirpath ] -> Fpath.of_string dirpath
+  | _ ->
+    Fmt.error_msg
+      "Expected one directory path in Smtzilla_utils.Sites.data, instead got: \
+       %d values"
+      (List.length Smtml_sites.Sites.data)
```
filipeom:
```suggestion
let smtzilla_data_dirpath () =
  match Smtml_sites.Sites.data with
  | [ dirp ] -> Fpath.of_string dirpath
  | _ ->
    Fmt.error_msg
      "Expected one directory path in Smtzilla_utils.Sites.data, instead got: \
       %d value"
      (List.length Smtml_sites.Sites.data)
```

Acho devia ser assim

PR #596, opened 2026-04-13
optimize Hashtbl.replace -> Hashtbl.add

redianthus:
This was showing up a little bit too much in my measurements (the one in `expr.ml` but I thought changing the others would do no harm).

EDIT: in the end I went with removing the Hashtbl, using a list is fine because 1. we never need to concatenate two lists, as we simply add one symbol at a time and 2. we can simply use `List.sort_uniq` at the end, which should be fine as there shouldn't be too many symbols.

### DIFF
--- a/src/smtml/expr.ml
+++ b/src/smtml/expr.ml
@@ -188,7 +188,7 @@ let get_symbols (hte : t list) =
     match view hte with
     | Val _ -> ()
     | Ptr { offset; _ } -> symbols offset
-    | Symbol s -> Hashtbl.replace tbl s ()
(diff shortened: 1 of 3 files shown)


filipeom (approved):
Is it because `Hashtbl.add` is faster in the happy path?

redianthus:
Yes, if you look at the implementation in the stdlib you'll see it is doing much less work (or maybe it's not in the happy path but simply becaus ewe have collisions and in this case, add is also better than replace).

filipeom (commented):
Oh cool thanks!

PR #603, opened 2026-04-13
make Expr.get_symbols_aux tail-rec

redianthus:
This is an experiment. This function is appearing too much in the profiling results to my taste. I made it tail-recursive by CPS transformation. I hope that TCO will reduce its cost.

If it does not fix the issue, I might go into defunctionalization. The code will likely be less readable so I'd like to try this one first. :sweat_smile:


filipeom:
What about memoizing the result of the function as well? It could help I think

redianthus:
Ah yes indeed, it's likely going to help too! :)
Will try that too

redianthus:
@filipeom, I think it's fine as is and we can merge.

filipeom (approved):
Nice, thanks!

PR #605, opened 2026-04-27
Explicitly depend on rresult

vbgl:
The file `src/smtml/feature_extraction.mli` refers to the module `Rresult` but does not explicitly link to it.

IIUC, the `Rresult` module is made available through the `bos` library (at version 0.2) but this no longer works with `bos` 0.3.0.

### DIFF
--- a/smtml.opam
+++ b/smtml.opam
@@ -36,6 +36,7 @@ depends: [
   "ppx_deriving"
   "ppx_enumerate"
   "prelude" {>= "0.5"}
+  "rresult"
   "scfg" {>= "0.5"}
   "mtime" {>= "2.0.0"}
   "yojson" {>= "1.6.0"}
--- a/dune-project
+++ b/dune-project
@@ -82,6 +82,7 @@
   ppx_enumerate
   (prelude
    (>= "0.5"))
+  rresult
   (scfg
    (>= "0.5"))
   (mtime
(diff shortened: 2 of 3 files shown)


smtml.opam
```diff
@@ -36,6 +36,7 @@ depends: [
   "ppx_deriving"
   "ppx_enumerate"
   "prelude" {>= "0.5"}
+  "rresult"
```
hra687261, line 39: You should add `rresult` to the `dune-project` file, the `smtml.opam` file is generated from it after a `dune build`. That's probably why the CI is broken.

hra687261 (changes requested):
Indeed, thanks! I think I'll eventually drop the dependency on `Rresult` as well.

filipeom (approved):
Thanks!

hra687261 (approved):

redianthus:
@filipeom we could add `(implicit_transitive_deps false)` in the `dune-project` file to avoid this kind of issue WDYT? (I thought it was there but apparently it's not)

filipeom:
> @filipeom we could add `(implicit_transitive_deps false)` in the `dune-project` file to avoid this kind of issue WDYT? (I thought it was there but apparently it's not)

Ah yes ofc! I always wanted this to be the default :) But I think it only works with ocaml 5 or something right? In any case I'll make a PR :smiley:

redianthus:
> But I think it only works with ocaml 5 or something right

I don't think so, it should be fine with any version IIRC

PR #609, opened 2026-04-28
Improve failure info in mappings

filipeom:


redianthus:
This is great but I feel like you could maybe avoid getting in the trouble of adding this everywhere by either running `OCAMLRUNPARAM=b whatever.exe` which should display a backtrace, or simply conditionally turn the backtrace recording on from the OCaml side (doing it unconditionally will hurt performances so we should avoid it).

Or is there something better with the approach in the current PR?

(In any case, I don't really mind, this was more a suggestion :))

filipeom:
It was mostly to avoid hardcoding stuff like: "Bitwuzla_mappings: Int.lt not implemented" because when I copy paste the failure strings I sometimes forget to update it and then it leads to some confusing messages :sweat_smile: But yeah, the backtrace is what I rely on mostly :smiley:

Issue #602, opened 2026-04-13
get_best_solver is costly


redianthus:
It takes 3% of the whole execution time on test-comp with a 30 seconds timeout.

hra687261:
Using strings as keys for the map of features is probably part of the problem. I'll change that soon.

hra687261:
Did not mean to close this ...

hra687261:
@redianthus do you have a simple way to reproduce the measure? are there command line options to pass to owi or symbocalypse to get these percentages or a test-comp run?

hra687261:
ping @redianthus

redianthus:
Oh sorry for the delay. No, it's not easy to reproduce. Basically, you need to run Test-Comp with perf and look at the flamegraph. You would start by recording with (you need to use a compiler with frame pointers enabled, and preferably flambda1 too) :

```shell-session
$ perf record -F 99 --call-graph fp,2000 -- dune exec -- symbocalypse testcomp owi 30 --solver=smtzilla --exploration=rarity-depth-aging
```

Then, here, in a normal situation you would go:

```shell-session
$ perf script report flamegraph
```

But it does not merge the various owi processes that are spawned by symbocalypse and it is thus unusable.
My first try was to use Brendan's scripts here: https://github.com/brendangregg/FlameGraph
And instead do:

```shell-session
$ perf script > out.perf
$ perl FlameGraph/stackcollapse-perf.pl out.perf | perl FlameGraph/flamegraph.pl > out.svg
```

This is very slow but it ultimately works (out.perf is 77Go...). So instead, you should open the recorded trace in hotspot and use the flamegraph visualization mode. The loading is slow but then it works well enough for me.

PR #611, opened 2026-04-30
Fix smtzilla python script and model

### LINKED ISSUE #598: smtzilla seems to always select Z3
cc @hra687261, when running on Test-Comp with a 30 seconds timeout on the 1600+ files, I see no trace of Bitwuzla anywhere! I checked, and it is available and installed.

It is correctly using smtzilla because I see calls to Feature_extraction.*.

hra687261:
The orignial script exported an incorrect model which is why only 1 solver was selected ... with this one, both z3 and bitwuzla are used when running owi on test comp.

The results are still not better than using z3 alone (779 vs 783 with 20s timeout), but fixing #602 and experimenting with different model parameters might help.

Closes https://github.com/formalsec/smtml/issues/598.

### DIFF
--- a/misc/model.json
+++ b/misc/model.json
@@ -1,784 +1,1450 @@
 {
-  "z3": {
-    "gradient_boost": false,
+  "bitwuzla": {
+    "gradient_boost": true,
     "model": {
-      "feature": "Lt",
-      "threshold": 13.5,
-      "left": {
-        "feature": "ConvertSI32",
-        "threshold": 0.5,
-        "left": {
-          "feature": "Eq",
+      "n_estimators": 5,
+      "init_value": 1132434.5224362633,
+      "trees": [
+        {
+          "feature": "PromoteF32",
           "threshold": 0.5,
           "left": {
-            "feature": "Ty_bitv",
-            "threshold": 0.5,
+            "feature": "Ty_fp",
+            "threshold": 20.5,
             "left": {
-              "feature": "Ty_fp",
-              "threshold": 0.5,
+              "feature": "ShrA",
+              "threshold": 23.0,
               "left": {
-                "feature": "Ty_regexp",
-                "threshold": -2.0,
+                "feature": "LtU",
+                "threshold": 109.0,
                 "left": {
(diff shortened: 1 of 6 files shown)


redianthus:
Thanks!

I'm not asking this for this PR but do you think it would makes sense to have something in smtml which can count how many times each solver was called and display the result, and then write a small test with two queries where we know one would go to Z3 and the other one to Bitwuzla, simply to make sure it keeps working?

> both z3 and bitwuzla are used when running owi on test comp

I'm curious how did you check this?

filipeom (approved):

filipeom:
> I'm not asking this for this PR but do you think it would makes sense to have something in smtml which can count how many times each solver was called and display the result, and then write a small test with two queries where we know one would go to Z3 and the other one to Bitwuzla, simply to make sure it keeps working?

I think it would be cool. We could have a histogram counter in the solver to track the amount of queries made for each. It would be cool

hra687261:
I added a cram test that shows different solvers are used for different queries. (I had to you use because smtml expects input streams to be in the smtml format, not smt2, which is not ideal imo, I don't think it's easy to check dynamically but maybe we can have it take smt2 by default and have a cmdline option for the smtml format?)
 
For counting how many times each solver was used it can be added, but as you said its unrelated to this PR so we can create a separate issue. It can probably be added to the stats map or something like that.

> I'm curious how did you check this?

Running owi on testcomp with a version of smtml that prints the selected solver. The debug cmdline options don't seem to be passed down to smtml. We could probably have a flag or simply set it by hand when owi is in verbose mode.

redianthus:
I think you can set `SMTML_VERBOSITY` maybe?

redianthus (approved):

hra687261:
> I think you can set `SMTML_VERBOSITY` maybe?

Is this predefined? I was not aware of it and it doesn't seem to be used in smtml.

redianthus:
Hum no sorry, there is `OWI_VERBOSITY` and I thought it was a thing provided by logs by default but it's not :sweat_smile:

PR #613, opened 2026-05-02
Restructure SMTZilla modules

### LINKED ISSUE #602: get_best_solver is costly

hra687261:
Some restructuring to prepare for some experimenting to fix #602

Due to the dependency (smtml -> smtzilla_utils) needed for the compile time loading of models, we can't "hide" the `feat` type, but now it is only seen in `feature` and `feature_map` which should make changing it easier.


filipeom (approved):
Thanks!

PR #629, opened 2026-06-08
Add indexed rotate operators

### LINKED ISSUE #625: Unexpected exception thrown
```ocaml

let n = Smtml.Typed.Bitv32.symbol (Smtml.Symbol.make (Smtml.Ty.Ty_bitv 32) "N0")

let one =  Smtml.Typed.Bitv32.one

let expr = Smtml.Typed.Bitv32.rotate_right one (Smtml.Typed.Bitv32.rotate_left n n )

module CVC5 = Smtml.Solver.Batch (Smtml.Cvc5_mappings)

let solver = CVC5.create ()

let () =
  match CVC5.check solver [ Smtml.Typed.Unsafe.unwrap  expr ] with
  | `Sat -> Format.printf "SAT@\n"
  | `Unsat -> Format.printf "UNSAT@\n"
  | `Unknown -> Format.printf "UNKNOWN@\n"
``` 

For the following example an exception is thrown.

```BASH
Fatal error: exception Invalid_argument("invalid kind 'BITVECTOR_ROTATE_LEFT', expected Terms with kind BITVECTOR_ROTATE_LEFT must have at least 1 children and at most 1 children (the one under construction has 2)")
Raised by primitive operation at Cvc5.Term.mk_term in file "cvc5.ml", line 98, characters 4-50
Called from Smtml__Mappings.Make.Make_.Encoder.encode_expr in file "src/smtml/mappings.ml", line 743, characters 16-33
Called from Smtml__Mappings.Make.Make_.Encoder.encode_expr in file "src/smtml/mappings.ml", line 742, characters 24-42
Called from Smtml__Mappings.Make.Make_.Encoder.encode_exprs.(fun) in file "src/smtml/mappings.ml", line 789, characters 27-44
Called from Stdlib__List.fold_left in file "list.ml", line 125, characters 24-34
Called from Smtml__Mappings.Make.Make_.Encoder.encode_exprs in file "src/smtml/mappings.ml", lines 787-791, characters 10-24
Called from Smtml__Mappings.Make.Make_.Solver.check in file "src/smtml/mappings.ml", line 951, characters 40-76
Called from Smtml__Utils.run_and_time_call in file "src/smtml/utils.ml", line 7, characters 15-19
Called from Dune__exe__Bug3 in file "smt/bug3.ml", line 12, characters 8-61
```

### LINKED ISSUE #627: update the signatures of Typed.Bitv32.rotate_left and right
We should change the function signatures to accept only constant. 
That will prevent invalid input types.

filipeom:
This is needed because some solvers may follow the SMT-LIB standard strictly and only implement the rotate operators using indicies.

Closes #625
Closes #627

BREAKING: This change intentionally breaks the `rotate_{left,right}` operators to take an integer as the first argument. This is to force users to make a decision on whether to continue using the previous behavior with `ext_rotate_{left,right}` or start using the new API.

### DIFF
--- a/src/smtml/bitwuzla_mappings.default.ml
+++ b/src/smtml/bitwuzla_mappings.default.ml
@@ -365,9 +365,17 @@ module Fresh_bitwuzla (B : Bitwuzla_cxx.S) : M = struct
 
     let rem_u t1 t2 = mk_term2 Kind.Bv_urem t1 t2
 
-    let rotate_left t1 t2 = mk_term2 Kind.Bv_rol t1 t2
+    let rotate_left n t =
+      let bitwidth = Sort.bv_size (Term.sort t) in
+      mk_term2 Kind.Bv_rol t (of_z (Z.of_int n) bitwidth)
 
-    let rotate_right t1 t2 = mk_term2 Kind.Bv_ror t1 t2
+    let rotate_right n t =
+      let bitwidth = Sort.bv_size (Term.sort t) in
+      mk_term2 Kind.Bv_ror t (of_z (Z.of_int n) bitwidth)
+
+    let ext_rotate_left t1 t2 = mk_term2 Kind.Bv_rol t1 t2
+
+    let ext_rotate_right t1 t2 = mk_term2 Kind.Bv_ror t1 t2
 
     let nego t = mk_term1 Kind.Bv_nego t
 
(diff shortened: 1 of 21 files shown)


redianthus:
Do you think it makes sense to try using `rotate` when `ext_rotate` is called on an constant value?

In the Z3 case, it might be faster, and in the others, it may allow to have a partial implementation in some cases?

filipeom:
> you think it makes sense to try using `rotate` when `ext_rotate` is called on an constant value?
> 
> In the Z3 case, it might be faster, and in the others, it may allow to have a partial implementation in some cases?

Ah good point! Yes I think it is totally worth it :smile:

redianthus:
@filipeom, I think we can merge this one?

filipeom:
Yes, I think so 😀

PR #630, opened 2026-06-08
SMTZilla: Use dedicated feat type instead of strings

hra687261:


filipeom (approved):

src/smtml/feature.ml
```diff
@@ -1,185 +1,185 @@
...
+  | Relop _ -> Relop_expr
+  | Cvtop _ -> Cvtop_expr
+  | Naryop _ -> Naryop_expr
+  | Extract _ -> Extract
+  | Concat _ -> Concat
+  | Binder _ -> Binder
 
-(* Define all constructors you want to track *)
-let ctor_names =
+let all_feats =
```
redianthus: I don't understand this part.

What is the reason to do:

```ocaml
let 1op = List.map of_1op [ ... ] in
let 2op = List.map of_2op [ ... ] in
let 3op = List.map of_3op [ ... ] in
1op @ 2op @ 3op
```

Instead of:

```ocaml
[ 1op1; 1op2; 2op1; 2op2; 3op1; 3op2 ]
```

?

(And then you might be able to use an array rather than a list, or even an immutable array!)

Said otherwise, can't you directly build an array of values of your new type?

redianthus (approved):

redianthus:
Looks good to me, there's only one thing I don't understand but it's not blocking

src/smtml/feature.ml
hra687261: Good remark! This was a quick change of the previous version, where of_xop converted an operator into a string, I had used it so that it does not silently cause problems if the string is changed in the of_xop function. I will change it.

hra687261: I did not know that it was possible to create immutable arrays with `[| ... |]`, on the other hand, it seems that they are only available since ocaml >= 5.4, so that might be a strong constraint and it might be better to use normal arrays.

PR #639, opened 2026-07-01
Use explicit result return in smtml parsing functions

filipeom:
It already bit me in the ass by not treating exceptions in injector. With explicit types I'm forced to treat them and also it allows me to use monadic syntax.

### DIFF
--- a/src/smtml/parse.ml
+++ b/src/smtml/parse.ml
@@ -2,8 +2,6 @@
 (* Copyright (C) 2023-2026 formalsec *)
 (* Written by the Smtml programmers *)
 
-exception Syntax_error of string
-
 module Smtml = struct
   open Lexer
   open Lexing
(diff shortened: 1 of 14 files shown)


src/smtml/parse.ml
```diff
@@ -14,15 +12,14 @@ module Smtml = struct
...
 
   let parse_with_error parser lexbuf =
-    try parser Lexer.token lexbuf with
-    | SyntaxError msg ->
-      raise (Syntax_error (Fmt.str "%a: %s" pp_pos lexbuf msg))
-    | Parser.Error ->
-      raise (Syntax_error (Fmt.str "%a: syntax error" pp_pos lexbuf))
+    try Ok (parser Lexer.token lexbuf) with
+    | SyntaxError msg -> Error (`Msg (Fmt.str "%a: %s" pp_pos lexbuf msg))
+    | Parser.Error -> Error (`Msg (Fmt.str "%a: syntax error" pp_pos lexbuf))
```
redianthus: You can use:

```ocaml
Fmt.error_msg "%a: syntax error" pp_pos lexbuf
```

which I found quite handy to get rid of all these parenthesis

src/smtml/parse.ml
```diff
@@ -59,16 +57,16 @@ module Smtlib = struct
...
-        (Syntax_error
+      Error
+        (`Msg
            (Fmt.str "%a: syntax error: %t" Dolmen.Std.Loc.print_compact loc msg)
         )
     | Dolmen.Std.Loc.Syntax_error (loc, `Advanced (x, _, _, _)) ->
-      raise
-        (Syntax_error
+      Error
+        (`Msg
```
redianthus: and probably here too

redianthus (approved):

src/smtml/parse.ml
filipeom: Indeed it makes formatting errors much better! Thanks :smile:

PR #643, opened 2026-07-02
fix issue 642 Unexpected exception thrown and fix pp_binop

amenahh:


src/smtml/eval.ml
```diff
@@ -597,6 +600,19 @@ module F32 = struct
     | Floor -> fp32_of_float @@ Float.floor f
     | Trunc -> fp32_of_float @@ Float.trunc f
     | Is_nan -> if Float.is_nan f then Value.True else Value.False
+    | Is_normal ->
+      if
+        Float.is_infinite f || Float.is_nan f || Float.equal Float.zero f
+        || is_subnormal f
+      then Value.False
+      else Value.True
```
filipeom:
```suggestion
      match Float.classify_float f with 
      | Float.FP_normal -> Value.True 
      | _ -> False
```

I feel this could be simplified to this

src/smtml/eval.ml
```diff
@@ -597,6 +600,19 @@ module F32 = struct
...
+    | Is_subnormal -> if is_subnormal f then Value.True else Value.False
+    | Is_negative ->
+      if Float.compare f Float.zero < 0 then Value.True else Value.False
+    | Is_positive ->
+      if Float.compare f Float.zero <= 0 then Value.True else Value.False
```
filipeom:
```suggestion
      if Float.compare f Float.zero >= 0 then Value.True else Value.False
```

Should be >= instead of <= right?

Also, not sure how `compare` handles comparisons with `-0` but let's hope it's sane 😅

src/smtml/eval.ml
```diff
@@ -597,6 +600,19 @@ module F32 = struct
...
+    | Is_infinite -> if Float.is_infinite f then Value.True else Value.False
+    | Is_zero -> if Float.equal Float.zero f then Value.True else Value.False
```
filipeom:
```suggestion
    | Is_zero -> begin
      match Float.classify_float f with 
      | Fp_zero -> Value.True
      | _ -> False
    end
```

My worry is here is because `0. = -0.` I think `Float.equal` handles this correctly, but this is more explicit

src/smtml/eval.ml
```diff
@@ -586,6 +586,9 @@ module F32 = struct
 
   let[@inline] neg x = Int32.logxor x Int32.min_int
 
+  let[@inline] is_subnormal f =
+    match Float.classify_float f with Float.FP_subnormal -> true | _ -> false
+
```
filipeom: Looks like this is no longer used. Can be removed

filipeom (approved):

filipeom:
Thanks for doing this! Much appreciated :smile:

PR #645, opened 2026-07-03
Add support for colibri2.0.6.0

filipeom:


dune-project
```diff
@@ -54,9 +54,9 @@
   (z3
    (< "4.16.0"))
   (colibri2
-   (<= "0.6.0"))
+   (< "0.6.0"))
```
redianthus, line 57: Is this correct? The PR title says adding support for 0.6.0 but this is excluding this specific version which makes me a little bit confused

redianthus: Oh sorry, you are actually excluding it from the *conflicts*, makes sense...

PR #652, opened 2026-07-20
complete SIMD

redianthus:
This is a dumb copy-paste from my concrete implementation in Owi, with adjustments until it compiles... There are probably wrong things in some places, but I think it's a good start. I think we can merge and let future testing find potential issues if you're OK with it @filipeom.

Also, note the concrete implementation in Owi is passing the whole reference test-suite. So, it's probably not that bad. :sweat_smile:

### DIFF
--- a/src/smtml/typed.ml
+++ b/src/smtml/typed.ml
@@ -456,6 +456,206 @@ module Bitv64 = struct
   let[@inline] extend_i32_u x = Expr.cvtop ty (Zero_extend 32) x
 end
 
+module Types_inner = struct
+  let int : int ty = Ty_int
+
+  let real : real ty = Ty_real
+
+  let regexp : regexp ty = Ty_regexp
+
+  let bool : bool ty = ty_bool
+
+  let string : string ty = Ty_str
+
+  let bitv8 : Bitv8.w ty = Bitv8.ty
+
+  let bitv16 : Bitv16.w ty = Bitv16.ty
+
+  let bitv32 : Bitv32.w ty = Bitv32.ty
+
+  let bitv64 : Bitv64.w ty = Bitv64.ty
+
+  let float32 : float32 ty = Ty_fp 32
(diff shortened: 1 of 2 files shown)


filipeom:
> This is a dumb copy-paste from my concrete implementation in Owi, with adjustments until it compiles... There are probably wrong things in some places, but I think it's a good start. I think we can merge and let future testing find potential issues if you're OK with it @filipeom.
> 
> Also, note the concrete implementation in Owi is passing the whole reference test-suite. So, it's probably not that bad. 😅

I skimmed the code and I think there might be one or two things that can be problematic. But it's not blocking. I would need to verify it manually later. I can probably try to do this in the weekend

filipeom (approved):

filipeom:
If you want to preserve the commit hash in Owi I can change PR policy to accept merge commit

redianthus:
> I skimmed the code and I think there might be one or two things that can be problematic. But it's not blocking. I would need to verify it manually later. I can probably try to do this in the weekend

Thanks, that would be nice if you can throw an eye on it. At the same time, Haskell is using the SIMD proposal when compiling to Wasm and we now have support for Haskell in Owi, so I expect to find a lot of issues this way. I was also thinking we could fuzz all of this with Amenah.

> If you want to preserve the commit hash in Owi I can change PR policy to accept merge commit

It's fine, I'll update it in Owi! :)

filipeom:
> Thanks, that would be nice if you can throw an eye on it. At the same time, Haskell is using the SIMD proposal when compiling to Wasm and we now have support for Haskell in Owi, so I expect to find a lot of issues this way. I was also thinking we could fuzz all of this with Amenah.

Yeah I think that will help a lot also. I just wanted to test some operators manually, that way I can confirm if my understanding of SIMD is correct while adding some tests and having more involvement in the code :smile:

PR #664, opened 2026-07-27
Release 0.30.0

filipeom:
Automated Release 0.30.0


filipeom:
I think we should wait for #660 before the next release. It's a major change but I think it doesn't affect anyone other than me?

redianthus:
Yes, I'm not using these I think.

PR #672, opened 2026-08-09
Add `ls` command to quickly list installed solvers and versions

filipeom:


filipeom:
@coderabbitai full review

PR #677, opened 2026-08-31
Fix model recovery from Alt-Ergo for negative values

hra687261:
For some reason (- x) is produced as (0 - x).


redianthus (approved):

hra687261:
Merging since this is pretty simple and does not affect the behavior of the rest

PR #678, opened 2026-08-31
FIx handling of uninterpreted function by Alt-Ergo and Colibri2

hra687261:
The idea is to simply store uninterpreted functions in the context (same as we do for uninterpreted symbols), so that we can declare for alt-ergo notably.

I encountered this issue while looking at an error that @stevendeo encountered while trying to use smtml with Alt-Ergo

### DIFF
--- a/src/smtml/altergo_mappings.default.ml
+++ b/src/smtml/altergo_mappings.default.ml
@@ -197,18 +197,24 @@ module M = struct
 
       let mk_model ~ctx model =
         AEL.ModelMap.fold
-          (fun ((hs, _, _) as id) g acc ->
-            let e = cgraph_to_value hs g in
-            let sym = aeid_to_sym id in
-            let tcst =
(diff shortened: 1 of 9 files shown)


redianthus (approved):

filipeom:
This looks like it's approved. Any other thing missing? I was playing with supporting lifting uninterpreted function models so It would be cool to test with these changes as well. Let me know.

redianthus:
I think we were waiting for you to confirm this can go in and merge?

hra687261:
> This looks like it's approved. Any other thing missing? I was playing with supporting lifting uninterpreted function models so It would be cool to test with these changes as well. Let me know.

Sorry for the late reply, I was thinking of trying to add at least support for applications of functions as its done for other solvers AFAIU, but I haven't gotten to it, so I think we can merge :)

PR #680, opened 2026-09-23
add a fuzzer

redianthus:
Hi @filipeom,

Here is the promised fuzzer! :D

Most of the work was done during @amenahh's internship.

I'd like to polish a few things, but I'm opening this as a draft already, it should not change that much.


filipeom:
This is so cool!

PR #681, opened 2026-09-23
Fix model generation for expressions with different return types

filipeom:
Model generation for expressions like `str.len "a"` would return `Str "1"` instead of `Int 1` because `value_of_term` would pass the type of the theory of the expression instead of the return type of the expression. This commit addresses it by introducing a new `Expr.return_type` function which correctly returns the return type of an expression instead of the theory of which the expression is encoded in.


filipeom:
@coderabbitai review

